PR #1, opened 2026-02-05
Eval command

ankrgyl:


tests/eval_fixtures.rs
```diff
@@ -0,0 +1,350 @@
...
+
+            let mut cmd = Command::new(&bt_path);
+            cmd.arg("eval");
+            if let Some(args) = config.args.as_ref() {
+                cmd.args(args);
+            }
+            if let Some(runner_cmd) =
+                resolve_runner(&dir, runner.as_deref(), python_runner.as_ref())
+            {
+                cmd.arg("--runner").arg(runner_cmd);
```
ibolmo: does the runner auto detect the right runtime? i did a simple detection in my branch (in sdk)

tests/eval_fixtures.rs
```diff
@@ -0,0 +1,350 @@
...
+            if let Some(tsx_path) = local_tsx_path(&dir) {
+                cmd.env("BT_EVAL_JS_RUNNER", tsx_path);
+            }
+
+            if let Some(python) = python_runner.as_ref() {
+                cmd.env("BT_EVAL_PYTHON_RUNNER", python);
+            }
+
+            let expect_success = config.expect_success.unwrap_or(true);
+            let status = cmd.status().expect("run bt eval");
```
ibolmo: using `--jsonl` i asserted the output had something interesting. should have documented it, but i found an example where the existing cli exited `0` but it was actually a failure.

PR #5, opened 2026-02-08
Add prompts command with list, view, and delete subcommands

parkerhendo:
## Summary
- Add `prompts` command with `list`, `view`, and `delete` subcommands
- Use comfy_table for consistent table formatting
- Add pager support for long output
- Make base arguments (api-url, org-name) globally available

## Test plan
- [ ] `braintrust prompts list` shows prompts in a table
- [ ] `braintrust prompts view <name>` opens prompt in web
- [ ] `braintrust prompts delete <name>` deletes after confirmation
- [ ] Table truncation handles emoji/unicode correctly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -195,6 +195,7 @@ dependencies = [
  "lingua",
  "open",
  "ratatui",
+ "regex",
  "reqwest",
  "serde",
  "serde_json 1.0.149",
--- a/src/prompts/api.rs
+++ b/src/prompts/api.rs
@@ -32,7 +32,11 @@ pub async fn list_prompts(client: &ApiClient, project: &str) -> Result<Vec<Promp
     Ok(list.objects)
 }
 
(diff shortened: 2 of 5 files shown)


ankrgyl (commented):
Thoughts from playing with it
 
* can we use `resolve_project(...)` so that there's a project selector instead of having to include one?
* I opened a prompt and it took me to

Ankurs-MacBook-Pro:~/projects/braintrust-cli-2 ankur$ bt prompts view --project prompt-test
✔ Select prompt · Testing 1234
✓ Opened https://staging-www.braintrust.dev/app/braintrustdata.com/p/prompt-test/prompts/ee6462eb-a0a6-445f-8dfc-08d215996c16 in browser

where does `staging-www.braintrust.dev` come from? i can't find it in the code.

* Also why does it instantly open a link in the UI? I'd rathe rsee the prompt definition in my CLI. I also think it's quite unintuitive for a coding agent that might try to run this

```
-n, --name <NAME_FLAG>   Prompt name (flag)
```

name is ambiguous. we should have slug here IMO

PR #15, opened 2026-02-12
fix bt eval with monorepo and without extensions in imports

ibolmo:
Fix 1: TypeScript extension resolution in installNodeModuleHooks (lines 399-472)

The resolve hook now wraps the next() call in a try/catch. When resolution fails for an extensionless relative specifier (like ./processors/tool-call-processor), it retries with .ts, .tsx, .mts, .cts extensions appended. This makes module.registerHooks() (Node 22.14+/23.5+) handle extensionless TS imports natively, so both require() and import() work without tsx hooks.

Fix 2: ESM-package-aware loading order in loadFiles (lines 821-920)

The old code always tried require() first for .ts files. On Node < 22.14, calling require() on a file in a "type": "module" package triggers require-of-ESM, which can corrupt the module graph and cause the "Unexpected status" internal assertion error (nodejs/node#54577).
The new code adds isInsideEsmPackage() to check the nearest package.json. For TS files in "type": "module" packages, it now tries import() first. This avoids the CJS/ESM interop bug on older Node versions while preserving the original require()-first behavior for CJS packages (where tsx's hooks work best).

Fix 3: if customers use parseArgs with strict will throw an exception because we passed files  in the args

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -284,8 +284,7 @@ async fn run_eval_files_once(
         EvalLanguage::JavaScript => build_js_command(runner_override, &js_runner, &files)?,
     };
 
-    let files_json =
-        serde_json::to_string(&files).context("failed to serialize eval file list")?;
+    let files_json = serde_json::to_string(&files).context("failed to serialize eval file list")?;
     cmd.env("BT_EVAL_FILES", files_json);
     cmd.envs(build_env(base));
     if no_send_logs {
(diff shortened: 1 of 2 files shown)


scripts/eval-runner.ts
```diff
@@ -191,6 +191,23 @@ function readRunnerConfig(): RunnerConfig {
...
+      !Array.isArray(parsed) ||
+      !parsed.every((f: unknown) => typeof f === "string")
+    ) {
+      throw new Error("BT_EVAL_FILES must be a JSON array of strings.");
+    }
+    return parsed;
+  }
+  // Fallback: read from argv for backwards compatibility (e.g., running
+  // the eval-runner directly outside of bt).
+  return process.argv.slice(2);
```
ankrgyl: what is this

PR #19, opened 2026-02-14
Login support

ankrgyl:
* Login with api keys or oauth
* Save profile configuration and the actual credentials in a keychain

### DIFF
--- a/src/login.rs
+++ b/src/login.rs
@@ -22,10 +22,11 @@ use serde::{Deserialize, Serialize};
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::net::TcpListener;
 
-use crate::{args::BaseArgs, ui};
+use crate::{
+    args::{BaseArgs, DEFAULT_API_URL, DEFAULT_APP_URL},
+    ui,
+};
 
-const DEFAULT_API_URL: &str = "https://api.braintrust.dev";
-const DEFAULT_APP_URL: &str = "https://www.braintrust.dev";
 const KEYCHAIN_SERVICE: &str = "com.braintrust.bt.cli";
 const OAUTH_SCOPE: &str = "mcp";
 const OAUTH_CALLBACK_TIMEOUT: Duration = Duration::from_secs(300);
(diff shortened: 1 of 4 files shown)


src/login.rs
```diff
@@ -1,45 +1,1817 @@
...
+    #[serde(default)]
+    oauth_client_id: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "snake_case")]
+enum AuthKind {
+    #[default]
+    ApiKey,
+    Oauth,
```
ankrgyl: capitalize A in OAuth?

src/login.rs
```diff
@@ -1,45 +1,1817 @@
...
+        println!("Auth source: BRAINTRUST_API_KEY environment variable");
+        if let Ok(profile_name) = std::env::var("BRAINTRUST_PROFILE") {
+            println!("Requested profile via BRAINTRUST_PROFILE={profile_name}");
+        } else if let Some(active_profile) = store.active_profile.as_deref() {
+            println!("Active saved profile (ignored while env key is set): {active_profile}");
+        }
+        return Ok(());
+    }
+
+    let selected_profile = std::env::var("BRAINTRUST_PROFILE")
```
ankrgyl: do not use std::env::var. Always use clap.

src/login.rs
```diff
@@ -1,45 +1,1817 @@
...
+
+    store.active_profile = None;
+    save_auth_store(&store)?;
+    println!("Cleared active profile.");
+    Ok(())
+}
+
+fn run_login_status() -> Result<()> {
+    let store = load_auth_store()?;
+    let env_api_key = std::env::var("BRAINTRUST_API_KEY")
```
ankrgyl: same. you cannot use std::env::var for anything that overlaps with CLI arguments

src/login.rs
```diff
@@ -1,45 +1,1817 @@
...
+    let token_url = format!("{}/oauth/token", api_url.trim_end_matches('/'));
+    let body = serde_json::json!({
+        "grant_type": "refresh_token",
+        "refresh_token": refresh_token,
+        "client_id": client_id,
+    });
+    exchange_oauth_token(&token_url, &body).await
+}
+
+async fn exchange_oauth_token(
```
ankrgyl: is there an oauth library we should be using?

PR #41, opened 2026-02-25
add parse args tests

ibolmo:
### TL;DR

Fixed issue where eval file paths in `process.argv` caused errors when user code used `parseArgs({ strict: true })`

### What changed?

Modified the eval runner to strip eval file paths from `process.argv` before loading user code, keeping only the first two elements (node executable and script name). Added a test case that uses `parseArgs` with strict mode to verify the fix works correctly.

### How to test?

Run the new `parse-args.eval.mjs` test file which calls `parseArgs({ strict: true })` - this would previously throw `ERR_PARSE_ARGS_UNEXPECTED_POSITIONAL` but should now pass without errors.

### Why make this change?

Customer reported that their eval files using `parseArgs` for command-line flags like `--shard` and `--experiment-name` were failing because the runner was leaking eval file paths into `process.argv` as unexpected positional arguments when strict parsing was enabled.

### DIFF
--- a/scripts/eval-runner.ts
+++ b/scripts/eval-runner.ts
@@ -708,8 +708,16 @@ function ensureBraintrustAvailable() {
   resolveBraintrustPath();
 }
 
+function getEvalFiles(): string[] {
+  const fromEnv = process.env.BT_EVAL_FILES_JSON;
+  if (fromEnv) {
+    return JSON.parse(fromEnv) as string[];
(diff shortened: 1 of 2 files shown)


parkerhendo (approved):
Pushed a fix if you want to validate on your end. Tests are now passing.

parkerhendo (approved):

PR #65, opened 2026-03-05
emit progress events from the python dev server

nselvidge:
## Summary        
  - Streaming eval results: Users watching eval runs in the dev server UI had no visibility into progress until the entire run completed. Now individual task results stream back in real-time.
  - Parent context support: Eval runs triggered from the app need their spans nested under the parent trace, so the dev server now accepts and propagates a parent span identifier.
  - Improved error reporting: When the eval runner crashed, the dev server returned a generic "exited with an error" message, making debugging difficult. Now stderr output is included.
  - Global state flush: The Python SDK sometimes dropped log events on process exit because the background flush hadn't completed. An explicit flush ensures all data reaches the API before the process terminates.
  - Musl build targets: Added x86_64-unknown-linux-musl and aarch64-unknown-linux-musl so the CLI can run on any Linux environment (e.g. Lambda, minimal containers) without glibc version dependencies.
  - SSE field naming: Aligned start event fields with the camelCase convention used by the rest of the API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,8 @@ targets = [
   "x86_64-apple-darwin",
   "aarch64-unknown-linux-gnu",
   "x86_64-unknown-linux-gnu",
+  "aarch64-unknown-linux-musl",
+  "x86_64-unknown-linux-musl",
   "x86_64-pc-windows-msvc",
 ]
 unix-archive = ".tar.gz"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -61,6 +61,8 @@ targets = [
   "x86_64-apple-darwin",
   "aarch64-unknown-linux-gnu",
   "x86_64-unknown-linux-gnu",
+  "aarch64-unknown-linux-musl",
+  "x86_64-unknown-linux-musl",
```
nselvidge: musl targets make this easier to run in lambda

parkerhendo (approved):
Tested it a fair bit locally and everything seems to be working as expected. Thanks for doing this!

PR #85, opened 2026-03-27
Cli install sdk evals

viadezo1er:
### bt setup instrument: interactive mode, language selection, and scoped permissions                                                                                        
                                                                                                                                                                           
Adds three new flags to bt setup instrument and wires them end-to-end through agent invocation and task generation.
                                                                                                                                                                           
```--interactive / -i``` opens the agent in its interactive TUI (Claude Code, etc.) so the user can review and approve each tool use.
```--yolo``` runs the agent in the background with bypassPermissions — no approval prompts.
```--language <LANG>``` restricts instrumentation to specific languages (python, typescript, go, java, ruby, csharp); repeatable; omit to let the agent auto-detect.                                                                                                                                      
                                                                  
#### Run-mode prompt (interactive terminal, no flags)

When none of the above flags are passed and the terminal is interactive, the user is asked how to run the agent. Background mode uses acceptEdits with --allowedTools scoped to the package managers for the selected language(s) only (e.g. uv for Python, npm/yarn/pnpm for TypeScript, dotnet for C#). Interactive TUI mode opens the agent's terminal UI.                                                                              
                                                                  
#### Language selection prompt

A multi-select prompt is shown between the workflow and run-mode prompts. Selecting "All languages" (the default) lets the agent auto-detect; selecting specific languages also narrows the background tool allowlist.


You'll notice since the mcp can be setup (and is the default option) when using bt setup, adding the mcp resources to this repo can make them redondant. However they are needed in case the mcp isn't setup.

### DIFF
--- a/src/setup/mod.rs
+++ b/src/setup/mod.rs
@@ -378,7 +378,7 @@ pub async fn run_setup_top(base: BaseArgs, args: SetupArgs) -> Result<()> {
     match args.command {
         Some(SetupSubcommand::Skills(setup)) => run_setup(base, setup).await,
         Some(SetupSubcommand::Instrument(instrument)) => {
-            run_instrument_setup(base, instrument).await
+            run_instrument_setup(base, instrument, false).await
         }
         Some(SetupSubcommand::Mcp(mcp)) => run_mcp_setup(base, mcp),
         Some(SetupSubcommand::Doctor(doctor)) => run_doctor(base, doctor),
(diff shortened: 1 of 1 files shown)


viadezo1er:
Needs rebasing (and deleting a helper file), and obviously the CI needs to pass.

viadezo1er:
The CI needs pinned SHA to run. Opening another PR for this.

PR #90, opened 2026-04-01
sign windows builds

ankrgyl:
https://linear.app/braintrustdata/issue/BT-4622/sign-window-builds-for-braintrust-cli

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Install dist
         shell: bash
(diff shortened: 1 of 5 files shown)


viadezo1er:
See https://axodotdev.github.io/cargo-dist/book/supplychain-security/signing/windows.html for how to sign the exe

viadezo1er (approved):
I think it works, but it needs to actually be tested in the real github action on a windows runner.

viadezo1er:
Running pre-release workflow to test if the test credentials work https://github.com/braintrustdata/bt/actions/runs/24801927430

viadezo1er (approved):
I'm not sure everything is needed but good to merge

PR #94, opened 2026-04-02
Cedric/better dx

viadezo1er:
```bt setup``` asks 0 and 2 questions: project and coding agent (and needs authentication)

#### org choice:
if --api-key is passed, use the provided api key
else if --prefer-profile, use the already registered profile (probably registered with OAuth previously)
else if the user has an api key in the env (variable BRAINTRUST_API_KEY), use it
else if there is already a registered profile, use it. If a profile is registered but the access token isn't accessible (for example storing the access token in plain text in ~/.config/bt/secrets.json but the file was deleted), just force re authentication
else same behavior as a user picking OAuth: give a url to paste in the browser

#### project choice:
if -p/--project is passed, use the specified project. If the project does not exist, stop the setup.
else if BRAINTRUST_DEFAULT_PROJECT exist in the env, same as for -p/--project
else the user has exactly one project, the default 'My Project' projects. Create a project in the user's organization named '{whoami}-test' where {whoami} is the result of the whoami command and use this project
else prompt the user to choose the project to use (creating a project is possible so the user isn't locked even with 0 projects)

#### skills, mcp, local, global:
by default, choose skills, no mcp and global scope
--no-skill, --no-mcp unselect skill/mcp
--local selects local instead of global
--no-mcp/skills don't remove pre-existing mcp/skills.

#### instrument:
by default, instrument
--(no-)instrument (un)select instrumentation
--no-instrument still configures mcp and skills, the only difference is that the agent isn't launched

#### agent:
if --agent AGENT flag is passed, use the specified agent
else if only one exists in [claude code, codex, cursor, opencode] (ie only one exist in PATH), use it
otherwise, prompt the user
If AGENT does not exist in PATH, try to launch it anyway.
Opencode support for starting with a prompt is lackluster as it can only start in the background.

#### workflow docs:
if one or multiple --workflow WORKFLOW flags are passed, fetch these workflows
otherwise, fetch all workflows

#### language:
try to identify the language based on file like pyproject.toml in the project root/sub folders (in current folder and subfolder with depth 1)
if nothing is found, alter the start prompt of the agent to find which language to instrument
conflicts with --no-instrument as the choice of language is only relevant if the agent is launched

#### TUI/background:
TUI by default
if --background is passed, launch the agent in the background
--tui exist only to conflict with --background (and --no-instrument)
Highly recommended to use --background with --yolo, as the agent won't be able to ask the user to allow it to run commands (and fail on step 1, checking if BRAINTRUST_API_KEY is in the env)

#### yolo mode
self explanatory for claude, codex, cursor
opencode is in yolo mode by default for the directory it is launched in
Besides, --yolo or equivalent don't exist in opencode yet (don't trust your LLM about this, there are many closed/merged/open PRs about this but just check ```opencode -h``` and you'll see it's not there).

#### interactive mode
-i/--interactive
Prompt the user to choose the configuration instead of passing options. Does not ask questions which are already answered by options or env variables.
```
$ bt setup -i
✔ Install scope · global (user-wide)
Set up coding-agent skills? yes
Set up MCP server? yes
Run instrumentation agent? yes
✔ Select agent to instrument this repo · claude
✔ Select additional workflow docs to prefetch (instrument is always included) · observe, annotate, evaluate, deploy
✔ How do you want to run the agent? · Interactive (TUI)
Grant agent full permissions? (bypass permission prompts) yes
```

### Bugs
--json is very buggy
workflows and scope broken

### Untested/poorly tested:
env variables
no-color, api/app-url, env-file, no-fetch-docs, refresh-docs, workers

### TODO
Better help message in the interactive wizard
better json objects, they are not consistent
only make the available agents selectable when multiple are on PATH, print the other ones but in grey or something
if in a folder, don't look at the parent folder when detecting language (both prompt and auto detect)

### CODE REVIEW

File: src/setup/mod.rs [hunk 1]
```diff
@@ -738,62 +987,185 @@ async fn select_project_with_skip(
     )
     .await?;
 
-    if projects.is_empty() {
-        bail!("no projects found in org '{}'", client.org_name());
+    // If the only project is the default "My Project" placeholder, auto-create one.
+    if projects.len() == 1 && projects[0].name == "My Project" {
```
AbhiPrasad: We should not auto-create, we should still always prompt the user for an action like this.

src/setup/mod.rs
```diff
@@ -681,55 +883,102 @@ async fn run_setup_wizard(mut base: BaseArgs, flags: WizardFlags) -> Result<()>
...
+    Ok(true)
 }
 
-async fn select_project_with_skip(
+/// Prompt the user to select a project from the org.
+///
+/// - If `--project` was specified, look it up and return it.
+/// - If the only project is the default "My Project" placeholder, auto-create `{whoami}-test`.
+/// - Otherwise, list all projects sorted alphabetically and let the user fuzzy-search and pick one.
+async fn select_project(
```
AbhiPrasad: This is very similar to `select_project_interactive`. Can we unify, or DRY somehow?

src/auth.rs
```diff
@@ -779,17 +778,7 @@ async fn run_login_oauth(base: &BaseArgs, args: AuthLoginArgs) -> Result<()> {
...
-    let selected = ui::fuzzy_select("Select login method", &methods, 0)?;
-
-    if selected == 0 {
-        login_interactive_oauth(base).await
-    } else {
-        login_interactive_api_key(base).await
-    }
-}
-
+#[allow(dead_code)]
```
AbhiPrasad: why allow dead code? Let's just remove if we are not using it.

ascii-logo-blue-small.txt
```diff
@@ -0,0 +1,5 @@
+[34m  ███  ███
+███      ███
+  ███  ███
+███      ███
+  ███  ███[0m
```
AbhiPrasad: This logo doesn't respect the `--no-color` option. We'll need to adjust accordingly.

skills/sdk-install/braintrust-url-formats.md
```diff
@@ -4,27 +4,28 @@
...
+| ------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| r                  | The root_span_id - identifies a trace                                                                                                                                            |
+| s                  | The span_id - identifies a specific span within the trace                                                                                                                        |
+| id                 | Legacy parameter for root_span_id in object URLs                                                                                                                                 |
+| BRAINTRUST_APP_URL | url where the app is accessible, www.braintrust.dev/app for non self-hosted instances. See the env variable BRAINTRUST_APP_URL 9and use www.braintrust.dev/app if it is not set) |
```
AbhiPrasad:
```suggestion
| BRAINTRUST_APP_URL | url where the app is accessible, www.braintrust.dev/app for non self-hosted instances. See the env variable BRAINTRUST_APP_URL and use www.braintrust.dev/app if it is not set) |
```

Cargo.toml
```diff
@@ -74,6 +74,7 @@ install-success-msg = ""
 
 [dev-dependencies]
 tempfile = "3"
+rexpect = "0.5"
```
AbhiPrasad: why add this?

src/setup/mod.rs
```diff
@@ -2409,6 +2993,23 @@ fn pick_agent_mode_target(candidates: &[Agent]) -> Option<Agent> {
...
+        agents.push(Agent::Claude);
+    }
+    if command_exists("cursor-agent") {
+        agents.push(Agent::Cursor);
+    }
+    if command_exists("opencode") {
+        agents.push(Agent::Opencode);
+    }
+    agents
+}
```
AbhiPrasad, line 2936: Can we cache this call?

src/setup/mod.rs
```diff
@@ -406,272 +468,412 @@ pub async fn run_setup_top(base: BaseArgs, args: SetupArgs) -> Result<()> {
...
+        tui: flag_tui,
+        background: flag_background,
+        agent: flag_agent,
         workflows: flag_workflows,
+        no_workflow: flag_no_workflow,
+        languages: flag_languages,
     } = flags;
+    const LOGO: &str = include_str!("../../ascii-logo-blue-small.txt");
+    eprintln!("{LOGO}");
+    eprintln!("\x1b[34mBraintrust\x1b[0m\n");
```
AbhiPrasad: we should use `dialoguer::console::style` here.

src/setup/mod.rs
AbhiPrasad, line 1: Ideally we can split this file up a bit, getting unwieldy to review

src/setup/mod.rs
```diff
@@ -738,62 +987,185 @@ async fn select_project_with_skip(
...
+    eprintln!(
+        "   {}",
+        style(format!("export BRAINTRUST_API_KEY={}", created.key)).dim()
+    );
+    eprintln!(
+        "   It is safe to cancel the setup now, export the key and restart the setup (with bt setup)."
+    );
+    eprintln!();
+    eprint!("   Press Enter to continue...");
+    let _ = std::io::stdin().lock().read_line(&mut String::new());
```
AbhiPrasad: We should check for `std::io::stdin().is_terminal` here (`is_interactive` helper)

src/setup/mod.rs
```diff
@@ -1683,36 +2166,55 @@ async fn run_agent_invocation(
 
             if *interactive {
                 // Inherit all streams so the user can interact with the agent directly.
+                #[cfg(unix)]
+                if !std::io::stdin().is_terminal() {
```
AbhiPrasad: This should use `is_interactive` helper.

AbhiPrasad (commented):
could we maybe split this PR up so the changes are a bit easier to review and land? Let me know if you need advice on how that should be. Claude told me this:

Based on the diff, here's how I'd split it into 4-5 independent PRs that can land in order:

### PR 1: Flag restructuring + deprecation
The mechanical rename/removal of CLI flags. Smallest blast radius, easiest to review.
- `--quiet` → `--verbose` (invert default)
- `--agents` (repeatable) → `--agent` (single), remove `AgentArg::All`
- `--no-mcp-skill` → `--no-skills` + `--no-mcp`
- Add `--no-instrument`, `--tui`/`--background`, `--no-workflow`
- Remove `--no-input`, move `--yes` to `#[arg(skip)]`
- `set_quiet(true)` as global default
- All the `BaseArgs` field renames (`quiet→verbose`) and test fixups
- Consider adding hidden deprecation aliases so existing scripts don't break

### PR 2: TTY fallback + agent detection
Self-contained infra changes that the wizard depends on.
- `tty_term()` function in `src/ui/select.rs` (`/dev/tty` fallback)
- `fuzzy_select` using `interact_on(&term)`
- `detect_runnable_agents()` (PATH-only check)
- `detect_agents` simplification (drop config-dir heuristics)
- `resolve_unambiguous_instrument_agent`
- `/dev/tty` stdin fallback in `run_agent_invocation` for interactive mode

### PR 3: Auth + API key creation
The OAuth-related changes, isolated from the wizard.
- `ensure_auth` rewrite (credential recovery flow, `is_missing_credential_error`)
- `maybe_create_api_key_for_oauth` (auto-create key, print-once)
- `login_interactive_oauth` made `pub(crate)`
- `#[allow(dead_code)]` on `login_interactive_api_key` (or just remove it)
- Remove `login_interactive` (the old method-chooser)
- Auth output tweaks (warning→info, spacing)

### PR 4: MCP key embedding
Depends on PR 3 (needs the API key to embed).
- `merge_mcp_config` takes `api_key` + `mcp_url` params
- `install_mcp_for_claude_user` (uses `claude mcp add -s user`)
- `mcp_url_from_api_url` helper
- `run_mcp_setup` becomes `async` (needs auth resolution)
- Updated tests for `merge_mcp_config` / `install_mcp_for_agent`

### PR 5: Wizard UX + template changes
The big UX PR, but now much smaller because infra landed in PRs 1-4.
- `--interactive`/`-i` flag and wizard flow
- ASCII logo + welcome banner
- `select_project` with auto-create `{whoami}-test` and `+ Create new project`
- Consolidate with `select_project_interactive` (deduplicate)
- `prompt_instrument_language_selection`, `prompt_instrument_workflow_selection` in wizard
- Agent invocation rework (Claude `--permission-mode`/`--settings`, Codex/Cursor TUI vs background split)
- `instrument-task.md` and `braintrust-url-formats.md` template changes
- `LanguageArg::doc_filename()`
- `render_instrument_task` rework
- SSE `strip_prefix` cleanup (could also go in PR 1)

---

**Dependency order**: PR 1 → PR 2 → PR 3 → PR 4 → PR 5 (though 2 and 3 are independent of each other and could land in parallel).

The biggest win is separating the flag restructuring (PR 1) from the wizard logic (PR 5) — those are the two hardest things to review when interleaved.

src/setup/mod.rs
viadezo1er: Removed this special case, now gives the option to choose the existing project or create a new one.

skills/sdk-install/braintrust-url-formats.md
viadezo1er: AZERTY/QWERTY muscle memory... It was supposed to be a '('.

src/setup/mod.rs
viadezo1er: Yes, unified

src/auth.rs
viadezo1er: Removed

ascii-logo-blue-small.txt
viadezo1er: done

src/setup/mod.rs
viadezo1er: Calling the function only once instead, so no need to cache now.

src/setup/mod.rs
viadezo1er: done

Cargo.toml
viadezo1er: I added (and forgot to commit) a small test file to check if ```bt setup --no-instrument``` would complete without any prompt with the correct env variables.
After re-reading it it's probably better to add it in a future PR.

src/setup/mod.rs
viadezo1er: done

src/setup/mod.rs
viadezo1er: done

src/ui/mod.rs
```diff
@@ -8,14 +8,9 @@ mod spinner;
...
 mod table;
 
-static NO_INPUT: AtomicBool = AtomicBool::new(false);
 static QUIET: AtomicBool = AtomicBool::new(false);
 static ANIMATIONS_ENABLED: AtomicBool = AtomicBool::new(true);
 
-pub fn set_no_input(val: bool) {
-    NO_INPUT.store(val, Ordering::Relaxed);
-}
-
```
parkerhendo, line 18: I don't know if we should remove this. there should be an explicit wait for users to tell the cli that the current session is not interactive/to not require user input

src/args.rs
```diff
@@ -44,10 +44,6 @@ pub struct BaseArgs {
     #[arg(long, global = true)]
     pub prefer_profile: bool,
 
-    /// Disable all interactive prompts
```
parkerhendo, line 47: see other comment. i don't think we should remove this, but it's possible I'm missing something.

src/ui/mod.rs
viadezo1er: You're correct, not removing it when splitting up this pr in 5 parts

PR #97, opened 2026-04-05
Add api and token commands

ankrgyl:


src/api.rs
```diff
@@ -0,0 +1,570 @@
...
+}
+
+#[derive(Debug, Serialize)]
+struct SpecOutput {
+    source_url: String,
+    operations: Vec<SpecOperation>,
+}
+
+async fn fetch_openapi_spec(source_url: &str) -> Result<String> {
+    let response = reqwest::Client::new()
```
nselvidge: nit: probably need DEFAULT_HTTP_TIMEOUT here

src/auth.rs
```diff
@@ -319,19 +346,129 @@ pub async fn run(base: BaseArgs, args: AuthArgs) -> Result<()> {
...
+    let token = auth.api_key.clone().ok_or_else(|| {
         anyhow::anyhow!(
             "no login credentials found; set BRAINTRUST_API_KEY, pass --api-key, or run `bt auth login`"
         )
     })?;
+    if !auth.is_oauth {
+        return Ok(ResolvedBearerToken {
+            token,
+            expires_at: None,
+        });
```
nselvidge, line 544: since this is used in `run_token` we'll log the permanent `BRAINTRUST_API_KEY` to stdout. I think that's fine but calling it out.

nselvidge (approved):
a couple nits but otherwise looks good

parkerhendo (approved):
command structure and such look good!

src/api.rs
ankrgyl: Fixed in bc32346. `fetch_openapi_spec` now uses `build_http_client(DEFAULT_HTTP_TIMEOUT)`, so it inherits the shared timeout, TLS root configuration, and user agent.

src/auth.rs
ankrgyl: Fixed in bc32346. `bt auth token` now exchanges API-key credentials through `/api/oauth/token` as well, so both OAuth and API-key profiles return a short-lived org-scoped token. Normal API-key authentication for other commands remains unchanged.

viadezo1er:
Context: https://github.com/braintrustdata/braintrust/commit/65ca09a83eda2b7a4dbeb8ab909154d49032b21e

PR #101, opened 2026-04-07
fix(setup): align braintrust skill name with install path

AbhiPrasad:
Rename the generated skill frontmatter from braintrust-cli to braintrust so installed skills match the parent directory name used by agent integrations.

Add a setup regression test to ensure future skill renders keep the expected frontmatter name and avoid skill conflict warnings.

pi was giving a warning for this:

```
[Skill conflicts]
   auto (user) ~/.agents/skills/braintrust/SKILL.md
     name "braintrust-cli" does not match parent directory "braintrust"
```

### DIFF
--- a/skills/shared/skill_frontmatter.md
+++ b/skills/shared/skill_frontmatter.md
@@ -1,5 +1,5 @@
 ---
-name: braintrust-cli
+name: braintrust
 version: 1.0.0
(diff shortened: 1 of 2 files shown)


viadezo1er (approved):
Thanks, I sometimes had issues with this

PR #104, opened 2026-04-09
Add `bt datasets` command for managing remote datasets

parkerhendo:
### TL;DR

Added a new `bt datasets` command for managing remote Braintrust datasets with CRUD operations and record upload support.

### What changed?

Added dataset management functionality:

- **New `bt datasets` command** with subcommands: `list`, `create`, `update`/`add`/`refresh`, `view`, and `delete`
- **Dataset operations** support multiple input methods: files (JSON/JSONL), inline JSON via `--rows`, or stdin
- **Record ID handling** with configurable ID fields via `--id-field`; `create` auto-generates UUID IDs when rows omit `id`, while `update`/`add`/`refresh` require explicit stable IDs
- **Direct row uploads** where `update`/`add`/`refresh` submit provided rows as merges without diffing remote rows first
- **Dataset previews** through `view`, using bounded row/value previews by default and `--full` when exact values are needed
- **Interactive features** including dataset selection, confirmation prompts, and browser opening
- **Shared API client usage** with BTQL pagination support and batch uploading
- **Updated CLI help** and README documentation with usage examples

The implementation includes a new `datasets` module with API client, record processing, upload handling, and individual command implementations. Added project context utilities for shared authentication and project resolution across commands.

### How to test?

```bash
# List datasets
bt datasets list

# Create datasets with different input methods
bt datasets create my-dataset --file records.jsonl
cat records.jsonl | bt datasets create my-dataset
bt datasets create my-dataset --rows "[{\"id\":\"case-1\",\"input\":{\"text\":\"hi\"},\"expected\":\"hello\"}]"

# Create can auto-generate UUID record IDs when rows omit id
bt datasets create my-dataset --rows "[{\"input\":{\"text\":\"hi\"},\"expected\":\"hello\"}]"

# Update existing datasets; add and refresh are visible aliases for update
bt datasets update my-dataset --file new-records.jsonl --id-field metadata.case_id
bt datasets add my-dataset --rows "[{\"id\":\"case-1\",\"input\":{\"text\":\"hi\"},\"expected\":\"hello\"}]"
bt datasets refresh my-dataset --file new-records.jsonl --id-field metadata.case_id

# View and delete datasets
bt datasets view my-dataset --verbose
bt datasets view my-dataset --json --full
bt datasets delete my-dataset --force
```

### Why make this change?

This change enables users to manage Braintrust datasets directly from the CLI without requiring local sync artifacts. It provides a streamlined workflow for dataset creation, updates, and inspection that integrates with existing Braintrust project management, making it easier to maintain evaluation datasets for AI/ML workflows.


ankrgyl:
* What if I want to upload records without an `id`? 
* If I upload something which contains extraneous fields, they get silently ignored:
```
Ankurs-MacBook-Pro:~/projects/braintrust-cli ankur$ cat test.json
{"foo": "bar", "id": 1}
```
just uploads the id

I think the PR description is a bit out of date. There's no `refresh` command, I think. But I also like that (I was about to ask, why would we have one!?)

parkerhendo:
> There's no refresh command, I think. But I also like that

`datasets refresh` exists! it's an alias for `datasets update` along with `datasets add`. Missed the update in the pr description.

viadezo1er:
Not done with the review yet, there's a linting error, `cargo fmt` fixes this.

src/datasets/api.rs
```diff
@@ -0,0 +1,248 @@
...
+            "fmt": "json",
+        });
+        let org_name = client.org_name();
+        let headers = if !org_name.is_empty() {
+            vec![("x-bt-org-name", org_name)]
+        } else {
+            Vec::new()
+        };
+        let response: DatasetRowsResponse =
+            client.post_with_headers("/btql", &body, &headers).await?;
```
ankrgyl: we should use the existing btql helpers we have.

src/datasets/api.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let query =
+            build_dataset_rows_query(dataset_id, MAX_DATASET_ROWS_PAGE_LIMIT, cursor.as_deref());
+        let body = serde_json::json!({
+            "query": query,
+            "fmt": "json",
+        });
+        let org_name = client.org_name();
+        let headers = if !org_name.is_empty() {
+            vec![("x-bt-org-name", org_name)]
```
ankrgyl: specifically, it's dangerous to reinvent logic like this

src/datasets/api.rs
```diff
@@ -0,0 +1,248 @@
...
+    let cursor_clause = cursor
+        .map(|cursor| format!(" | cursor: {}", btql_quote(cursor)))
+        .unwrap_or_default();
+    format!(
+        "select: * | from: dataset({}) | filter: created >= {} | limit: {}{}",
+        sql_quote(dataset_id),
+        btql_quote(DATASET_ROWS_SINCE),
+        limit,
+        cursor_clause
+    )
```
ankrgyl: no string formatting to construct queries.

src/datasets/api.rs
```diff
@@ -0,0 +1,248 @@
...
+    let path = format!("/v1/dataset/{}", encode(dataset_id));
+    client.delete(&path).await
+}
+
+fn build_dataset_rows_query(dataset_id: &str, limit: usize, cursor: Option<&str>) -> String {
+    let cursor_clause = cursor
+        .map(|cursor| format!(" | cursor: {}", btql_quote(cursor)))
+        .unwrap_or_default();
+    format!(
+        "select: * | from: dataset({}) | filter: created >= {} | limit: {}{}",
```
ankrgyl: unless you are downloading the dataset to get exact values out, you should use a `preview_length` so that we don't accidentally download too much data.

src/datasets/api.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+fn btql_quote(value: &str) -> String {
+    serde_json::to_string(value)
+        .unwrap_or_else(|_| format!("\"{}\"", value.replace('\\', "\\\\").replace('"', "\\\"")))
+}
+
+fn sql_quote(value: &str) -> String {
+    format!("'{}'", value.replace('\'', "''"))
+}
```
ankrgyl: these functions should not exist. we should construct the parse tree.

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+pub(crate) const DATASET_UPLOAD_BATCH_SIZE: usize = 1000;
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct PreparedDatasetRecord {
+    pub id: String,
+    pub input: Option<Value>,
+    pub expected: Option<Value>,
+    pub metadata: Option<Map<String, Value>>,
+    pub tags: Option<Vec<String>>,
+}
```
ankrgyl, line 39: can we reuse this type from the SDK (if it exists there)?

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+    let Some(contents) = read_input_contents(input_path, inline_rows)? else {
+        return Ok(None);
+    };
+    if contents.trim().is_empty() && input_path.is_none() && inline_rows.is_none() {
+        return Ok(None);
+    }
+    Ok(Some(parse_record_objects(&contents)?))
+}
+
+fn parse_record_objects(contents: &str) -> Result<Vec<Map<String, Value>>> {
```
ankrgyl, line 122: it would be better to use serde_json's streaming parser if it's a series of `{ ... }` objects (like jsonl). in general, we should avoid having this parsing logic if we can.

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+    Ok(records)
+}
+
+fn validate_supported_fields(
+    object: &Map<String, Value>,
+    id_root: &str,
+    row_index: usize,
+) -> Result<()> {
+    const BASE_ALLOWED_FIELDS: [&str; 6] =
+        ["id", "input", "expected", "output", "metadata", "tags"];
```
ankrgyl: is it possible to deduplicate this with the struct definition? Maybe we can use serde to validate that the object itself contains no other fields?

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+    Ok(PreparedDatasetRecord {
+        id,
+        input: object.get("input").cloned(),
+        expected: object
+            .get("expected")
+            .cloned()
+            .or_else(|| object.get("output").cloned()),
+        metadata: parse_metadata(object.get("metadata"))?,
+        tags: parse_tags(object.get("tags"))?,
+    })
```
ankrgyl, line 302: this logic should be done via serde

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+    let path = id_field
+        .split('.')
+        .map(str::trim)
+        .filter(|part| !part.is_empty())
+        .map(ToOwned::to_owned)
+        .collect::<Vec<_>>();
+    if path.is_empty() {
+        bail!("id field path cannot be empty");
+    }
+    Ok(path)
```
ankrgyl, line 344: this looks dangerous (what if there are `.` in the path names).

ankrgyl, line 344: this looks dangerous (what if there are `.` in the path names).

ankrgyl: also make sure to unit test it, or ideally, it shouldn't exist.

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+}
+
+fn lookup_object_path<'a>(object: &'a Map<String, Value>, path: &[String]) -> Option<&'a Value> {
```
ankrgyl: we should have an existing helper for this. if we don't, put this in a utility file so that other commands can use it

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+    let mut current = object.get(path.first()?.as_str())?;
+    for part in path.iter().skip(1) {
+        current = current.as_object()?.get(part.as_str())?;
+    }
+    Some(current)
+}
+
+fn coerce_id_value(value: &Value) -> Result<String> {
```
ankrgyl: what is the difference between just serializing this as a string and then trimming it? Also why are we trimming it in the first place?

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+
+fn parse_metadata(value: Option<&Value>) -> Result<Option<Map<String, Value>>> {
+    match value {
+        None | Some(Value::Null) => Ok(None),
+        Some(Value::Object(metadata)) => Ok(Some(metadata.clone())),
+        Some(_) => bail!("dataset record metadata must be a JSON object"),
+    }
+}
```
ankrgyl: this should be enforced via struct definition/serde

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+
+fn parse_tags(value: Option<&Value>) -> Result<Option<Vec<String>>> {
```
ankrgyl: this should be enforced via struct definition/serde

src/datasets/records.rs
```diff
@@ -0,0 +1,474 @@
...
+                Value::String(value) => Ok(value.clone()),
+                _ => bail!("dataset record tags[{index}] must be a string"),
+            })
+            .collect::<Result<Vec<_>>>()
+            .map(Some),
+        Some(_) => bail!("dataset record tags must be an array of strings"),
+    }
+}
+
+fn generated_record_id(object: &Map<String, Value>, row_index: usize) -> Result<String> {
```
ankrgyl: i don't think we should hash contents to produce default ids. We should just use uuidv4(). Because the point of having a stable id is so that if you change the contents, you can replace it. So there's no semantic meaning to this.

ankrgyl: You can look at how this is done in the JS SDK https://github.com/braintrustdata/braintrust-sdk-javascript as a reference

ankrgyl: Fixed in this branch: default dataset record IDs now come from a shared CLI helper `crate::utils::new_uuid_id()` backed by UUID v4, and dataset record handling no longer hashes row contents for generated IDs.

I kept the `sha2` dependency because `src/sync.rs` still uses it.

Verified with:
- `cargo fmt --check`
- `cargo test datasets`
- `cargo test utils::ids`

src/datasets/api.rs
ankrgyl: Fixed in this branch by routing dataset row BTQL fetches through the shared `ApiClient::btql` helper instead of constructing the `/btql` body and org header locally.

I also extended `BtqlResponse<T>` with an optional `cursor` so the dataset pagination path can keep using cursor responses through the shared helper.

Verified with:
- `cargo fmt --check`
- `cargo test datasets`
- `cargo test btql_response`

src/datasets/api.rs
ankrgyl: Fixed the duplicated request/response plumbing here as part of the BTQL helper cleanup: dataset row listing now calls `ApiClient::btql::<DatasetRow>(&query)`, and the dataset-local `/btql` body/header construction plus `DatasetRowsResponse` are gone.

The remaining query string construction / quoting concerns are still tracked separately under the parse-tree/string-formatting comments.

Verified with:
- `cargo fmt --check`
- `cargo test datasets`
- `cargo test btql_response`

src/datasets/api.rs
ankrgyl: Fixed. Dataset row listing now includes `preview_length` by default for `bt datasets view`, including `--json` and `--verbose`, so values are previewed unless the user explicitly asks for exact values.

I also added `bt datasets view --full` / `BT_DATASETS_VIEW_FULL`, which switches the BTQL query to `preview_length: -1`. The view output now points users to `bt sync pull dataset:<id>` when they need the full dataset exported to files.

src/datasets/api.rs
ankrgyl: Fixed. Dataset row listing now builds a structured BTQL query instead of formatting a query string. I added a small reusable Rust BTQL AST in `src/btql.rs`, added `ApiClient::btql_structured(...)`, and converted `build_dataset_rows_query()` to produce the parse tree with `dataset(<literal>)`, the required `created >= "1970-01-01T00:00:00Z"` filter, `preview_length`, `limit`, and `cursor`.

src/datasets/api.rs
ankrgyl: Fixed as part of the structured BTQL cleanup. `src/datasets/api.rs` no longer has local `btql_quote()` / `sql_quote()` helpers for this query path; dataset ids, timestamps, and cursors are passed as structured BTQL literals and serialized by serde.

src/datasets/records.rs
ankrgyl: Checked the Rust SDK types. `braintrust_sdk_rust::DatasetRecord` is not a good fit for this CLI upload input boundary because it requires `id: String` while `bt datasets create` can omit ids, and it does not deny unknown fields. I added a local serde input shape instead, matching the current dataset upload fields (`id`, `input`, `expected`, `metadata`, `tags`), with `id` typed as `Option<String>`. `output` is rejected as an unknown upload field.

src/datasets/records.rs
ankrgyl: Fixed. Standard `id` is now deserialized as `Option<String>`, and custom `--id-field` values go through the same string-only parser. We no longer trim ids or coerce numbers/bools to strings; empty, null, and non-string ids are rejected with tests.

src/datasets/api.rs
ankrgyl: Follow-up after review: I removed the reusable typed BTQL AST module. The dataset query is still structured, but it is now built directly as a `serde_json::Value` with `json!({ ... })` for this query pattern.

src/datasets/records.rs
ankrgyl: Follow-up for SDK alignment: update/add/refresh uploads now mark rows with `_is_merge: true`, while create-time seed uploads omit `_is_merge`, matching the SDK insert/update split. Added focused coverage around the upload row shape.

src/datasets/records.rs
ankrgyl: Fixed. Record input parsing now uses a single streaming parser pass with `serde_json::Deserializer::from_str(...).into_iter::<Value>()`. Plain JSONL and adjacent `{...}{...}` object streams are treated as record streams. A single top-level array or `{ "rows": [...] }` object is expanded only when it is the whole stream, preserving array and `bt datasets view --json` wrapper support.

src/datasets/records.rs
ankrgyl: Kept dot syntax per follow-up, but replaced the naive splitter with an explicit parser. Literal dots are escaped as `\.` and literal backslashes as `\\`; malformed paths like empty segments, dangling escapes, or unknown escapes now fail. Added CLI/README docs and unit coverage for literal dots/backslashes.

ankrgyl: Kept dot syntax per follow-up, but replaced the naive splitter with an explicit parser. Literal dots are escaped as `\.` and literal backslashes as `\\`; malformed paths like empty segments, dangling escapes, or unknown escapes now fail. Added CLI/README docs and unit coverage for literal dots/backslashes.

src/datasets/records.rs
ankrgyl: Kept dot syntax per follow-up, but replaced the naive splitter with an explicit parser. Literal dots are escaped as `\.` and literal backslashes as `\\`; malformed paths like empty segments, dangling escapes, or unknown escapes now fail. Added CLI/README docs and unit coverage for literal dots/backslashes.

ankrgyl: Follow-up for this helper-specific thread: the object traversal helper now lives in `src/utils/json_object.rs` and `datasets/records.rs` calls through that shared utility. Added focused utility tests for nested lookup, missing paths, non-object intermediates, empty paths, and literal-dot path segments. Verified with `cargo test utils::json_object` and `cargo test datasets::records`.

ankrgyl:
PR body follow-up: updated the description and test examples so `update`/`add`/`refresh` are described accurately. It now calls out that `refresh` is a visible alias for `update`, `create` can generate UUID record ids when missing, `update`/`add`/`refresh` require explicit stable ids and submit merge rows directly, and `view` uses previews by default with `--full` for exact values. Also verified generated CLI help shows `update` with `[aliases: add, refresh]` and examples for both aliases.

ankrgyl:
Follow-up commit review item resolved: `bt datasets view` no longer uses `select: *` for dataset rows. The BTQL query now selects only the row fields we want to expose and round-trip (`id`, `input`, `expected`, `metadata`, `tags`, `origin`). Upload validation no longer strips `dataset_id` / `created`; system fields remain invalid row fields. Fixtures now assert that `view --json` output omits system upload fields like `dataset_id` / `_is_merge` while preserving `origin`.

ankrgyl:
Follow-up commit review item resolved: documented the `{ "rows": [...] }` input shape as wrapper handling for `bt datasets view --json` compatibility. Sibling fields on the wrapper are ignored, but row-level validation remains strict after extraction; added a focused test where an unknown field inside a wrapped row is rejected.

ankrgyl:
Follow-up commit review item resolved: kept the dataset row page-size cap behavior and preserved coverage for default page size, remaining-row cap, and stopping at the requested limit. Added a fixture case where the mock BTQL server returns more rows than requested and the CLI still emits only the requested row count with `rows_truncated: true`.

src/datasets/records.rs
ankrgyl: Follow-up after the later row-selection change: the local CLI input shape now accepts `origin` as well (`id`, `input`, `expected`, `metadata`, `tags`, `origin`). The Rust SDK type still is not reused here for the same reasons: upload create can omit `id`, and the CLI boundary needs strict unknown-field validation.

src/datasets/records.rs
ankrgyl: Follow-up correction: the earlier note about stripping `dataset_id` / `created` is no longer current. `bt datasets view` now selects only upload-compatible/view-useful row fields (`id`, `input`, `expected`, `metadata`, `tags`, `origin`), and upload validation no longer has a system-field strip exception; system fields remain invalid row fields.

ankrgyl (approved):

PR #105, opened 2026-04-10
Flag restructuring for bt setup

viadezo1er:
Mechanical rename/removal of CLI flags:
- `--quiet` → `--verbose` (invert default)
- `--agents` (repeatable) → `--agent` (single), remove `AgentArg::All`
- `--no-mcp-skill` → `--no-skills` + `--no-mcp`
- Add `--no-instrument`, `--tui`/`--background`, `--no-workflow`
- Move `--yes` to `#[arg(skip)]`
- ```--interactive/-i``` flag, that more or less preserves the previous behavior of asking all questions

Part 1 of https://github.com/braintrustdata/bt/pull/94

### DIFF
--- a/src/setup/mod.rs
+++ b/src/setup/mod.rs
@@ -53,21 +53,58 @@ pub struct SetupArgs {
     #[command(subcommand)]
     command: Option<SetupSubcommand>,
 
-    /// Set up coding-agent skills (skips interactive selection in wizard)
-    #[arg(long)]
+    /// Set up coding-agent skills [default]
+    #[arg(long, conflicts_with = "no_skills")]
     skills: bool,
 
-    /// Set up MCP server (skips interactive selection in wizard)
-    #[arg(long)]
+    /// Do not set up coding-agent skills
+    #[arg(long, conflicts_with = "skills")]
+    no_skills: bool,
+
+    /// Set up MCP server
+    #[arg(long, conflicts_with = "no_mcp")]
     mcp: bool,
 
-    /// Run instrumentation agent (skips interactive prompt in wizard)
+    /// Do not set up MCP server [default]
+    #[arg(long, conflicts_with = "mcp")]
+    no_mcp: bool,
+
+    /// Run instrumentation agent [default]
     #[arg(long)]
     instrument: bool,
 
-    /// Skip skills and MCP setup (skips interactive selection in wizard)
-    #[arg(long, conflicts_with_all = ["skills", "mcp"])]
-    no_mcp_skill: bool,
+    /// Do not run instrumentation agent (skills and MCP are still configured)
+    #[arg(
+        long,
+        conflicts_with = "instrument",
+        conflicts_with = "tui",
+        conflicts_with = "background",
+        conflicts_with = "yolo"
+    )]
+    no_instrument: bool,
+
+    /// Run the agent in interactive TUI mode [default]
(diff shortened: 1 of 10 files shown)


viadezo1er:
pre-commit run --all-files does not show any issue locally...

src/setup/mod.rs
```diff
@@ -680,50 +792,153 @@ async fn run_setup_wizard(mut base: BaseArgs, flags: WizardFlags) -> Result<()>
...
+        .map(|profile| profile.name.clone())
+        .ok_or_else(|| anyhow!("no auth profiles found"))
 }
 
-async fn select_project_with_skip(
+fn should_force_reauth(err: &anyhow::Error) -> bool {
+    let msg = err.to_string();
+    msg.contains("oauth refresh token missing")
+        || msg.contains("oauth profile requested but none selected")
+        || msg.contains("re-run `bt auth login --oauth")
```
AbhiPrasad: Can we use an actual error here? Instead of a string match?

src/setup/mod.rs
```diff
@@ -738,21 +953,74 @@ async fn select_project_with_skip(
...
+    name: &str,
+) -> Result<crate::projects::api::Project> {
+    if let Some(project) = crate::projects::api::get_project_by_name(client, name).await? {
+        return Ok(project);
     }
+    match crate::projects::api::create_project(client, name).await {
+        Ok(project) => Ok(project),
+        Err(_) => crate::projects::api::get_project_by_name(client, name)
+            .await?
+            .ok_or_else(|| anyhow!("failed to create project '{name}'")),
```
AbhiPrasad: this discards the original error. We should log or chain with the original error.

src/setup/mod.rs
```diff
@@ -738,21 +953,74 @@ async fn select_project_with_skip(
     )
     .await?;
 
-    if projects.is_empty() {
-        bail!("no projects found in org '{}'", client.org_name());
```
AbhiPrasad, line 742: We should still have a `projects.is_empty` check before `projects.len() == 1`.

src/setup/mod.rs
```diff
@@ -2354,38 +2507,31 @@ fn resolve_scope_from_flags(
...
+    prompt: &str,
+    allow_prompt: bool,
+) -> Result<Agent> {
+    if let Some(arg) = requested {
+        return Ok(map_agent_arg(arg));
     }
 
-    if requested.contains(&AgentArg::All) {
-        return ALL_AGENTS.to_vec();
+    let path_agents = detected_agents_on_path(detected);
```
AbhiPrasad, line 2599: We need a `path_agents.is_empty` check here (I guess also checking for `allow_prompt`?)

src/setup/mod.rs
```diff
@@ -2409,6 +2573,16 @@ fn pick_agent_mode_target(candidates: &[Agent]) -> Option<Agent> {
     candidates.first().copied()
 }
 
+fn detected_agents_on_path(detected: &[DetectionSignal]) -> Vec<Agent> {
+    let mut agents = BTreeSet::new();
+    for signal in detected {
+        if signal.reason.contains("binary found in PATH") {
```
AbhiPrasad: string assert is bad, we should match on an error value or enum value directly.

AbhiPrasad (approved):

PR #109, opened 2026-04-13
TTY fallback for bt setup and agent detection

viadezo1er:
Self-contained infra changes that the wizard depends on.
- `tty_term()` function in `src/ui/select.rs` (`/dev/tty` fallback)
- `fuzzy_select` using `interact_on(&term)`
- `detect_runnable_agents()` (PATH-only check)
- `detect_agents` simplification (drop config-dir heuristics)
- `resolve_unambiguous_instrument_agent`
- `/dev/tty` stdin fallback in `run_agent_invocation` for interactive mode

Cursor skill path detection, global quiet mode are fixed in the [3rd PR](https://github.com/braintrustdata/bt/pull/110).

PATH only detection of the agent seem ok, I don't think people would want to install skills/mcp without having the agent.

Part 2 of https://github.com/braintrustdata/bt/pull/94

### DIFF
--- a/src/setup/mod.rs
+++ b/src/setup/mod.rs
@@ -1278,21 +1278,32 @@ async fn run_instrument_setup(
     print_hint: bool,
 ) -> Result<()> {
     let home = home_dir().ok_or_else(|| anyhow!("failed to resolve HOME/USERPROFILE"))?;
-    let root = find_git_root().ok_or_else(|| {
-        anyhow!(
-            "instrument setup requires running inside a git repository (could not find .git in parent chain)"
-        )
-    })?;
+    let root = find_git_root()
+        .map(Ok)
+        .unwrap_or_else(|| std::env::current_dir().context("failed to get current directory"))?;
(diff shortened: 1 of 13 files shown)


src/setup/mod.rs
```diff
@@ -3733,4 +4179,52 @@ mod tests {
...
+        let old = std::env::current_dir().expect("cwd");
+        std::env::set_current_dir(&root).expect("cd root");
+        let detected = detect_agents();
+        std::env::set_current_dir(old).expect("restore cwd");
+
+        assert!(
+            detected.iter().all(|signal| signal.on_path),
+            "detect_agents should only emit PATH-based signals"
+        );
+    }
```
AbhiPrasad, line 4200:
```suggestion
    #[test]
    fn detect_agents_reports_only_path_signals() {
       let detected = detect_agents();
       assert!(
           detected.iter().all(|signal| signal.on_path),
           "detect_agents should only emit PATH-based signals"
       );
     }
```

AbhiPrasad (approved):

PR #110, opened 2026-04-13
Auth + API key creation when using bt setup

viadezo1er:
The OAuth-related changes, isolated from the wizard.
- `ensure_auth` rewrite (credential recovery flow, `is_missing_credential_error`)
- `maybe_create_api_key_for_oauth` (auto-create key, print-once)
- `login_interactive_oauth` made `pub(crate)`
- `#[allow(dead_code)]` on `login_interactive_api_key` (or just remove it)
- Remove `login_interactive` (the old method-chooser)
- Auth output tweaks (warning→info, spacing)

Part 3 of https://github.com/braintrustdata/bt/pull/94

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -13,8 +13,8 @@ use braintrust_sdk_rust::{BraintrustClient, LoginState};
 use clap::{Args, Subcommand};
 use crossterm::event::{self, Event, KeyCode, KeyEventKind};
 use dialoguer::{Confirm, Input, Password};
-use oauth2::basic::{BasicClient, BasicTokenType};
-use oauth2::reqwest::async_http_client;
+use oauth2::basic::{BasicClient, BasicErrorResponseType, BasicRequestTokenError, BasicTokenType};
+use oauth2::reqwest::{async_http_client, AsyncHttpClientError};
 use oauth2::{
     AuthUrl, AuthorizationCode, ClientId, CsrfToken, EmptyExtraTokenFields, PkceCodeChallenge,
     PkceCodeVerifier, RedirectUrl, RefreshToken, Scope, StandardTokenResponse, TokenResponse,
@@ -623,7 +623,8 @@ pub async fn resolve_auth(base: &BaseArgs) -> Result<ResolvedAuth> {
             ),
         )
     })?;
-    let refreshed = refresh_oauth_access_token(&api_url, &refresh_token, client_id).await?;
+    let refreshed =
+        refresh_oauth_access_token(&api_url, &refresh_token, client_id, &profile_name).await?;
     save_profile_oauth_access_token(&profile_name, &refreshed.access_token)?;
     if let Some(next_refresh_token) = refreshed.refresh_token.as_ref() {
         if next_refresh_token != &refresh_token {
@@ -1152,7 +1153,9 @@ async fn run_login_refresh(base: &BaseArgs) -> Result<()> {
         println!("Cached access token expiry before refresh: unknown");
     }
 
-    let refreshed = refresh_oauth_access_token(&api_url, &refresh_token, &client_id).await?;
+    let refreshed =
+        refresh_oauth_access_token(&api_url, &refresh_token, &client_id, profile_name.as_str())
+            .await?;
     save_profile_oauth_access_token(profile_name.as_str(), &refreshed.access_token)?;
     let mut refresh_rotated = false;
     if let Some(next_refresh_token) = refreshed.refresh_token.as_ref() {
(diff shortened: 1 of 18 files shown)


src/auth.rs
```diff
@@ -831,53 +889,7 @@ async fn run_login_oauth(base: &BaseArgs, args: AuthLoginArgs) -> Result<()> {
...
-        base.app_url.clone(),
-        selected_org.as_ref().map(|org| org.name.clone()),
-    )?;
-
-    base.profile = Some(profile_name.clone());
-    Ok(profile_name)
-}
-
-async fn login_interactive_oauth(base: &mut BaseArgs) -> Result<String> {
+pub(crate) async fn login_interactive_oauth(base: &mut BaseArgs) -> Result<String> {
```
AbhiPrasad, line 967: This needs a `ui::is_interactive` check

src/auth.rs
```diff
@@ -1631,6 +1633,7 @@ fn select_login_org(
         }
     }));
     let label_refs: Vec<&str> = labels.iter().map(String::as_str).collect();
+    println!("\n\nA Braintrust organization is usually a team or a company.");
```
AbhiPrasad: we should use `eprintln!` instead

src/setup/mod.rs
```diff
@@ -1243,6 +1528,7 @@ fn prompt_instrument_workflow_selection() -> Result<Option<Vec<WorkflowArg>>> {
     }))
 }
 
+#[allow(dead_code)]
```
AbhiPrasad: we can remove `#[allow(dead_code)]`

AbhiPrasad (approved):

viadezo1er:
## Auth spec
For case where the user should be prompted but ```bt setup``` runs in non interactive mode (--no-input, --json), stop the setup.
-o/--org/-p/--project/--profile/'~/.bt/config.json' (config.json configures org and project fields)and their respective env variables: flags > env variables > config file in case of conflict between 2 indicators of the same field (like --org and BRAINTRUST_ORG_NAME).
When populating a profile , use the same behavior as if 'bt auth login' was run. If the profile to use is not specified:
- For OAuth login, if a profile with the same auth_kind, api_url, app_url, org_name, user_name and email already exist (profile name does not matter), just update the oauth_client_id, oauth_access_expires_at and the stored secret without prompting the user.
   - If there are multiple OAuth profiles matching org/user, update the one that has the largest oauth_access_expires_at (ie the most recently created/updated).
- For API key login, if a profile with the same auth_kind, api_url, and org_name already exist (profile name does not matter), create a new profile with the new api_key_hint and stored secret.

When trying to login with OAuth, first try to use the cached OAuth token, if it's missing/stale and there's a valid refresh token use it to get a new OAuth token, else browser-based OAuth workflow.
When a variable should be checked but there is an error (for example, --profile incorrect-profile-name), the setup fails and print an error.
A removed/revoked API key is treated as if it does not match any org.
In case of network error, 5xx from Braintrust, wrong API call, print an error and stop.
To check if a project is in an org, list the projects of the org with
```bash
curl --request GET \
  --url https://api.braintrust.dev/v1/project \
  --header 'Authorization: Bearer <token>'
```
then search the returned project list for the requested project name.

'goto aaa' means 'continue the program the line after the [aaa] line'.
Like in python, if there is no else statement just continue to the next line.
'If XXX is passed' means 'if XXX is in flags or env variables or config file'.
Decision tree:
```
if need authentication and instrumenting:
  All authentication use cases need an API key
  if API key passed:
    if profile passed:
      goto profile-passed-1
    else if org passed:
      if prefer-profile passed and 0 of the available profile's org matches the passed org:
        fail, end
      if API key's org matches passed org:
        [project-passed-1]
        if project is passed:
          if the project is in the passed org:
            use API key, end
          else:
            fail, end
        else:
          use API key, end
      else:
        OAuth, create API key, goto project-passed-1
    else if prefer-profile passed:
      if the number of available profile is 0:
        use API key, add a profile as if bt auth login with API key, end
      else if 1:
        if API key's org and profile's org match:
          use API key, end
        else:
          OAuth, create API key, end
      else if number of profiles' org matching the API key's org is 0:
        OAuth, create API key, end
      else if >=1:
        use API key, end
  goto no-org-no-profile-prefer-profile
end

[no-org-no-profile-prefer-profile]
if API key passed:
  if project is in API key's org:
    use API key, end
OAuth, create API key, end

[profile-passed-1]
if org passed:
  [profile-passed-2]
  if org and profile's org match:
    if API key passed:
      if the API key's org matches the profile's org:
        if project is passed and does not exist in org:
          fail, end
        else:
          use API key, end
    OAuth login.
    if project is passed and is not in the passed org:
      fail, end
    else:
      create API key, end
  else:
    fail, end
else:
  consider that passed org is the same as the profile's org
  goto profile-passed-2
```

PR #115, opened 2026-04-15
MCP key embedding

viadezo1er:
Many of the commits are due to rebasing.

- `merge_mcp_config` takes `api_key` + `mcp_url` params
- `install_mcp_for_claude_user` (uses `claude mcp add -s user`)
- `mcp_url_from_api_url` helper
- `run_mcp_setup` becomes `async` (needs auth resolution)
- Updated tests for `merge_mcp_config` / `install_mcp_for_agent`

Mcp configuration needs the relevant agent in PATH.
`bt setup --mcp --agent codex` will fail if codex does not exist.

PR4 of https://github.com/braintrustdata/bt/pull/94


src/setup/mod.rs
```diff
@@ -3487,41 +3511,192 @@ fn install_mcp_for_agent(
...
+            InstallScope::Local => install_mcp_for_codex_local(local_root, home, api_key, mcp_url),
+            InstallScope::Global => install_mcp_for_codex(mcp_url, api_key),
+        },
+        Agent::Cursor => install_mcp_for_cursor(scope, local_root, home, api_key, mcp_url),
+        Agent::Opencode => install_mcp_for_opencode(scope, local_root, home, api_key, mcp_url),
+    }
+}
+
+fn install_mcp_for_codex(mcp_url: &str, api_key: &str) -> Result<AgentInstallResult> {
+    std::env::set_var("BRAINTRUST_API_KEY", api_key);
```
AbhiPrasad: We don't need to mutate the entire env here, we can remove this. We already do `.env("BRAINTRUST_API_KEY", api_key)` and ditto below.

src/ui/select.rs
```diff
@@ -61,25 +61,37 @@ pub fn fuzzy_select_opt<T: ToString>(
...
-        .unwrap_or(0);
-
+        .unwrap_or(if projects.is_empty() { 0 } else { 1 });
     let label = select_label.unwrap_or("Select project");
     let selection = fuzzy_select(label, &names, default)?;
-    Ok(projects[selection].name.clone())
+
+    if selection == 0 {
+        let name: String = Input::with_theme(&ColorfulTheme::default())
+            .with_prompt("New project name")
```
AbhiPrasad: we don't want this to appear in every usage of `select_project`. For example this helper is used by `bt projects delete`. We need to make sure that `api::create_project` is really only there for setup/init.

src/setup/mod.rs
```diff
@@ -3487,41 +3511,192 @@ fn install_mcp_for_agent(
...
+            "mcp",
+            "add",
+            "-s",
+            scope_name,
+            "--transport",
+            "http",
+            "braintrust",
+            mcp_url,
+            "--header",
+            &format!("Authorization: Bearer {api_key}"),
```
AbhiPrasad, line 3623: Let's make sure none of these commands end up writing the api key to `.mcp.json` or similar. Or atleast we need to be careful before writing to a file that might be commited in git.

src/setup/mod.rs
```diff
@@ -504,6 +510,11 @@ async fn run_setup_wizard(mut base: BaseArgs, flags: WizardFlags) -> Result<()>
         no_workflow: flag_no_workflow,
         languages: flag_languages,
     } = flags;
+    const LOGO: &str = include_str!("../../ascii-logo-small.txt");
+    eprintln!("{LOGO}");
+    eprintln!("\x1b[34mBraintrust");
+    eprintln!("Eval everything\x1b[0m\n");
```
AbhiPrasad: needs to use `console::style`

src/setup/mod.rs
```diff
@@ -3487,41 +3511,192 @@ fn install_mcp_for_agent(
...
+}
+
+fn install_mcp_for_codex_local(
+    local_root: Option<&Path>,
+    home: &Path,
+    api_key: &str,
+    mcp_url: &str,
+) -> Result<AgentInstallResult> {
+    let root = scope_root(InstallScope::Local, local_root, home)?;
+    let path = root.join(".codex/config.toml");
```
AbhiPrasad, line 3540: is this a thing for codex? idk if codex allows for local config like this.

viadezo1er: It does, it's not that well documented but:
_you can also scope MCP servers to a project with .codex/config.toml_
in https://developers.openai.com/codex/mcp (and it works).

src/setup/mod.rs
viadezo1er: Claude writes the API key to ```.mcp.json``` (project) or ```~/.claude.json``` (user)
Cursor to ```(~/).cursor/.mcp.json```
Opencode to ```(~/.config/opencode/)opencode.json```

So the API key will get stored.

I'd really like to say that nothing ever gets added/commited/pushed by error buuuttt

Should we add these files to the ```.gitignore``` when setting up an mcp?

AbhiPrasad: okay you know what, let's tackle this afterwards. I'm not sure what the best decision here is, and we probably want to take a look to see what other similar tools.

src/setup/mod.rs
viadezo1er: Done

src/ui/select.rs
viadezo1er: Done

src/setup/mod.rs
viadezo1er: Done

AbhiPrasad (approved):

PR #116, opened 2026-04-16
test(status): add strip_ansi helper for ANSI code removal

parkerhendo:
strip ansi characters from output in tests before asserting so tests pass properly.


viadezo1er (approved):
Good to merge

PR #118, opened 2026-04-17
Dotenv not loaded by default

viadezo1er:
2 bug fixes:
auth-related bt commands (bt auth, bt switch for example) now require explicit --env-file or BRAINTRUST_ENV_FILE to look at .env files

OAuth login was no longer an option in `bt auth login` since b02d8b82e5eef8b3afd740cad6be19ea5eace81f

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -784,6 +784,16 @@ async fn run_login_set(base: &BaseArgs, args: AuthLoginArgs) -> Result<()> {
     if args.oauth {
         return run_login_oauth(base, args).await;
     }
+
+    let has_explicit_api_key = base.api_key.as_ref().is_some_and(|k| !k.trim().is_empty());
+    if !has_explicit_api_key && ui::can_prompt() {
+        let methods = ["OAuth (browser)", "API key"];
+        let selected = ui::fuzzy_select("Select login method", &methods, 0)?;
+        if selected == 0 {
+            return run_login_oauth(base, args).await;
+        }
+    }
+
     let interactive = ui::can_prompt();
 
     let api_key = match base.api_key.clone() {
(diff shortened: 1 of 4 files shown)


src/switch.rs
```diff
@@ -479,4 +507,52 @@ mod tests {
...
+            true,
+            true,
+            || panic!("org picker should not be called"),
+            &mut interactive,
+        )
+        .expect("resolve");
+
+        assert_eq!(profile, None);
+        assert!(interactive);
+    }
```
parkerhendo, line 557: This looks identical to `resolve_profile_for_switch_skips_org_prompt_when_api_key_infers_profile`?

src/switch.rs
```diff
@@ -232,6 +235,31 @@ fn apply_switch_config(cfg: &mut config::Config, org_name: &str, project: &api::
     cfg.project_id = Some(project.id.clone());
 }
 
+fn resolve_profile_for_switch<F>(
+    has_api_key_override: bool,
+    prompting_for_project_only: bool,
+    is_interactive: bool,
+    select_profile_interactive: F,
+    interactive: &mut bool,
```
parkerhendo, line 243: Nit: This function could return `Result<(Option<String>, bool)>` instead of using `&mut bool` to pass the value back to the caller so both outputs are in the return type. Would be easier to trace than a hidden mutation.

src/auth.rs
```diff
@@ -784,6 +784,16 @@ async fn run_login_set(base: &BaseArgs, args: AuthLoginArgs) -> Result<()> {
     if args.oauth {
         return run_login_oauth(base, args).await;
     }
+
+    let has_explicit_api_key = base.api_key.as_ref().is_some_and(|k| !k.trim().is_empty());
```
parkerhendo, line 788: Nit: This `base.api_key...` also exists in `switch.rs`. Could we move it to be a method on BaseArgs?

parkerhendo (approved):
A couple nits, but otherwise OAuth fix and env loading changes look correct to me.

src/switch.rs
viadezo1er: Removed

src/auth.rs
viadezo1er: Done

src/switch.rs
viadezo1er: Done

PR #132, opened 2026-04-20
feat: support for gemini coding agent

viadezo1er:
`gemini /path/to/file.md` while the current folder is not a trusted folder in `~/.gemini/trustedFolders.json` will still run the prompt. This is not our responsibility but is worth noting. It also means `bt setup --agent gemini` will start instrumentation regardless of the status (trusted/not trusted) of the current folder.

`~$ gemini mcp add ...` fails. Since `bt setup` uses `gemini mcp add ...` to add the mcp to gemini, the mcp won't be installed if `bt setup` is ran in `~`.

Like claude, cursor and opencode, gemini stores the BRAINTRUST api key in plain text.

Also removed the `codex mcp remove` call just before `codex mcp add`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,7 +319,7 @@ Current behavior:
 - Use `--refresh-docs` in setup (or `bt docs fetch --refresh`) to clear old docs before re-fetching.
 - `cursor` is local-only in this flow. If selected with `--global`, `bt` prints a warning and continues installing the other selected agents.
 - Claude integration installs the Braintrust skill file under `.claude/skills/braintrust/SKILL.md`.
-- Gemini integration installs `.gemini/skills` as an alias to the shared `.agents/skills/braintrust/SKILL.md` skill path.
+- Gemini integration symlinks `.gemini/skills` to `.agents/skills/braintrust/SKILL.md`.
 - Cursor integration installs `.cursor/rules/braintrust.mdc` with the same shared Braintrust guidance plus an auto-generated command-reference excerpt from this README.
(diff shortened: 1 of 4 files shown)


viadezo1er:
ah have a bug with background gemini

README.md
```diff
@@ -319,6 +319,7 @@ Current behavior:
 - Use `--refresh-docs` in setup (or `bt docs fetch --refresh`) to clear old docs before re-fetching.
 - `cursor` is local-only in this flow. If selected with `--global`, `bt` prints a warning and continues installing the other selected agents.
 - Claude integration installs the Braintrust skill file under `.claude/skills/braintrust/SKILL.md`.
+- Gemini integration installs `.gemini/skills` as an alias to the shared `.agents/skills/braintrust/SKILL.md` skill path.
```
AbhiPrasad: we should really say symlink. I think the alias function name should be renamed, but we can do that in a follow up.

AbhiPrasad (approved):
Can you expand on why you removed `codex mcp remove` in the PR description?

viadezo1er:
> Can you expand on why you removed `codex mcp remove` in the PR description?

Slop clean-up. What's said in the pr description + running multiple `codex mcp add braintrust-test --url 'https://api.braintrust.dev/mcp' --bearer-token-env-var BRAINTRUST_API_KEY` is fine so cleaning previous mcp installation is not necessary.

(I suppose you meant 'Can you expand on `codex mcp remove` in the PR description?')

README.md
viadezo1er: Done

PR #134, opened 2026-04-21
Add param argument to CLI

nselvidge:
Adds the ability to set a parameter value from CLI arguments. 

taking an incremental approach to implementing params to keep diffs small and easy to review, but planning to implement this full spec: https://www.notion.so/braintrustdata/Local-Parameters-348f7858028980b9b825f3f4e4b6bc76

### DIFF
--- a/scripts/eval-runner.py
+++ b/scripts/eval-runner.py
@@ -65,6 +65,7 @@ class RunnerConfig:
     sample_seed: int | None
     dev_mode: str | None
     dev_request_json: str | None
+    params: dict[str, Any] | None
 
 
 @dataclass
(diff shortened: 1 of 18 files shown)


scripts/eval-runner.py
```diff
@@ -1016,6 +1030,9 @@ async def run_single_evaluator(
             return evaluator_instance, None, None, err
 
         progress_cb = create_progress_reporter(sse, evaluator_instance.evaluator.eval_name)
+        if config.params is not None and evaluator_instance.evaluator.parameters is not None:
+            evaluator = evaluator_instance.evaluator
+            evaluator.parameters = validate_parameters(config.params, evaluator.parameters)
```
parkerhendo: I might be wrong here, but I think this check being outside the `try` block on L1036 allows validate_parameters(...) exceptions to escape run_single_evaluator instead of being returned as the normal per-evaluator error tuple.

parkerhendo (approved):
One thing we may want to address, but otherwise seems good!

scripts/eval-runner.py
nselvidge: Thanks Parker, good callout. I updated the code to filter out any params that aren't defined in the schema before validating. 

There's still an edge case where two evals with the same parameter key but a different type schema will throw, but I think that's OK personally. 

Added some tests to verify the behavior.

PR #135, opened 2026-04-22
Inform user API key is created during bt setup

viadezo1er:
Users who don't have an API key in their env (for example new users) need one to run the instrumented code.
So after OAuth login, when needed, an API key is created.
However it wasn't shown to the user unless `bt setup` was run with the `-v/--verbose` flag.
Now this is printed unless `-q/--quiet` is passed.
```
! Created Braintrust API key 'cedric@braintrustdata.com-created-by-bt-setup34' for instrumentation and exported it to this setup process:

   sk-...

   To reuse it later in your shell, run:
   export BRAINTRUST_API_KEY=sk-...
```


AbhiPrasad (approved):
nice, good call

PR #138, opened 2026-04-23
Compile bt for Windows ARM64

viadezo1er:
`bt.exe` now has an ARM64 version.

Cross compilation of the ring crate to Windows aarch64 fails, so the compilation of the ARM64 bt.exe runs on Windows server ARM64.
Compilation of the auto-updater on Windows server ARM64 fails, so is has been removed (this settings is global and can'tr be turned off just for one architecture). `bt self update` exists anyway so it's not an issue.

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -140,6 +140,14 @@ jobs:
             echo "$HOME/.cargo/bin" >> $GITHUB_PATH
           fi
 
+      - name: Install sha256sum (macOS ARM64)
+        if: ${{ runner.os == 'macOS' && runner.arch == 'ARM64' }}
+        shell: bash
+        run: |
+          printf '#!/bin/sh\nexec shasum -a 256 "$@"\n' > /tmp/sha256sum
+          chmod +x /tmp/sha256sum
+          echo /tmp >> "$GITHUB_PATH"
+
       - name: Install dist
         run: ${{ matrix.install_dist.run }}
 
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -113,7 +113,6 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       BUILD_MANIFEST_NAME: target/distrib/${{ join(matrix.targets, '-') }}-dist-manifest.json
-      HAS_SSLDOTCOM_SIGNING: ${{ secrets.SSLDOTCOM_USERNAME != '' && secrets.SSLDOTCOM_PASSWORD != '' && secrets.SSLDOTCOM_CREDENTIAL_ID != '' && secrets.SSLDOTCOM_TOTP_SECRET != '' }}
     steps:
       - name: Enable windows longpaths
         run: git config --global core.longpaths true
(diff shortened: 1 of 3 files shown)


viadezo1er:
I'll push force, too much was removed by dist.

viadezo1er:
```toml
[dist.github-custom-runners]
aarch64-pc-windows-msvc = "windows-11-arm"
```
makes the compilation of the Windows aarch64 version happen in an ARM64 Windows server.

dist-workspace.toml
```diff
@@ -3,6 +3,8 @@ members = ["cargo:."]
 
 # Config for 'dist'
 [dist]
+# Skip checking whether the specified configuration files are up to date
+allow-dirty = ["ci"]
```
AbhiPrasad, line 7: intentional change?

viadezo1er: Yes otherwise this happens https://github.com/braintrustdata/bt/actions/runs/24857419412/job/72773867253?pr=138

viadezo1er:
Force push as I reverted the wrong commit

viadezo1er:
Code signing works for the sandbox.

dist-workspace.toml
```diff
@@ -11,21 +13,20 @@ ci = "github"
...
+[dist.github-custom-runners]
+aarch64-pc-windows-msvc = "windows-11-arm"
 # The archive format to use for non-windows builds (defaults .tar.xz)
 unix-archive = ".tar.gz"
 # The archive format to use for windows builds (defaults .zip)
 windows-archive = ".zip"
-# Path that installers should place binaries in
-install-path = ["$XDG_BIN_HOME/", "$XDG_DATA_HOME/../bin", "~/.local/bin"]
-# Whether to install an updater program
-install-updater = true
```
parkerhendo, line 29: Does removing this break the `self update` command?

viadezo1er: No, [`This standalone program will be installed as the name yourpackage-update`](https://axodotdev.github.io/cargo-dist/book/installers/updater.html) so it removes `bt-update`. `bt self update` is defined in `src/self_update.rs`.

viadezo1er: And install path still exists in `dist-workspace.toml`

viadezo1er:
[Doc says](https://axodotdev.github.io/cargo-dist/book/supplychain-security/signing/windows.html):
```toml
[workspace.metadata.dist]
ssldotcom-windows-sign = "prod"  # or "test" if you are using a sandbox account
```
Once we get the EV certificate, change `test` to `prod` in `dist-workspace.toml`.

Edit: Changed my mind, let's release this now so bt.exe is available for aarch64 now, signing with a real certificate will come later.

.github/workflows/release-canary.yml
AbhiPrasad, line 1: do we need to add `!fromJson(needs.plan.outputs.val).announcement_is_prerelease` here like we do with `release.yml`?

.github/workflows/release.yml
```diff
@@ -139,7 +147,7 @@ jobs:
         run: ${{ matrix.packages_install }}
 
       - name: Configure SSL.com signing env
-        if: ${{ runner.os == 'Windows' && env.HAS_SSLDOTCOM_SIGNING == 'true' }}
+        if: ${{ contains(toJSON(matrix.targets), 'windows') && env.HAS_SSLDOTCOM_SIGNING == 'true' && !fromJson(needs.plan.outputs.val).announcement_is_prerelease }}
```
AbhiPrasad: we can keep `runner.os == 'Windows'`, we don't need the `toJSON(matrix.targets)` check

dist-workspace.toml
```diff
@@ -11,21 +13,20 @@ ci = "github"
...
 installers = ["shell", "powershell", "homebrew"]
 homepage = "https://github.com/braintrustdata/bt"
 # Target platforms to build apps for (Rust target-triple syntax)
-targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc"]
+targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "aarch64-pc-windows-msvc", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc"]
+# Path that installers should place binaries in
+install-path = ["$XDG_BIN_HOME/", "$XDG_DATA_HOME/../bin", "~/.local/bin"]
+
+[dist.github-custom-runners]
+aarch64-pc-windows-msvc = "windows-11-arm"
```
AbhiPrasad: This needs to go below `unix-archive` and `windows-archive`, otherwise those fields don't map to `[dist]` in the toml

.github/workflows/release.yml
viadezo1er: I erroneously thought windows compilation for x86-64 happened on Linux... I probably got confused when I was cross-compiling from Mac to Windows aarch64.
Should have red https://axodotdev.github.io/cargo-dist/book/ci/customizing.html that does say default is windows-2022

.github/workflows/release-canary.yml
viadezo1er: Rather I forgot to remove code signing for canary release (don't see the point of signing the canary release, considering the amount of bugs we often have in them and our current shipping speed)

dist-workspace.toml
viadezo1er: thanks for the catch

AbhiPrasad (approved):

PR #152, opened 2026-04-25
Cedric/add qwen coding agent

viadezo1er:
Feat: support Qwen in `bt setup`
Chore: refactor agent functions to reuse code


viadezo1er:
Found some bugs and corrected them in https://github.com/braintrustdata/bt/pull/156, no point in keeping this open

PR #157, opened 2026-04-28
fix(setup): clarify default project selection

AbhiPrasad:
fixes https://linear.app/braintrustdata/issue/BT-4949/when-using-the-setup-script-right-after-creating-a-fresh-org-just

Update the setup project selector so users in a fresh org can either create an intentional project immediately or continue with the starter project. Instead of showing a bare "My Project" entry, the interactive selector now renders:

```
+ Create new project
My Project (default starter project)
```

The annotation is part of the displayed fuzzy-select label, so users can match it by typing terms like "default" or "starter". Selecting "+ Create new project" prompts for a project name and creates it. Selecting "My Project (default starter project)" uses the existing starter project. Other setup flows that only allow existing projects are unchanged.

### DIFF
--- a/src/ui/select.rs
+++ b/src/ui/select.rs
@@ -16,7 +16,7 @@ use crate::{
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ProjectSelectMode {
     ExistingOnly,
-    AllowCreate,
+    AllowCreateWithDefaultProjectNote,
 }
 
 /// Open a Term for interactive prompts.
@@ -82,7 +82,6 @@ pub async fn select_project(
     mode: ProjectSelectMode,
 ) -> Result<api::Project> {
     let mut projects = with_spinner("Loading projects...", api::list_projects(client)).await?;
-
     projects.sort_by(|a, b| a.name.cmp(&b.name));
 
     let names = project_selection_labels(&projects, mode);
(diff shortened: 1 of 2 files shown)


src/ui/select.rs
```diff
@@ -115,9 +114,17 @@ pub async fn select_project(
...
 fn project_selection_labels(projects: &[api::Project], mode: ProjectSelectMode) -> Vec<String> {
-    if matches!(mode, ProjectSelectMode::AllowCreate) {
+    if mode_allows_create(mode) {
         let mut labels = vec!["+ Create new project".to_string()];
-        labels.extend(projects.iter().map(|project| project.name.clone()));
+        labels.extend(projects.iter().map(|project| {
+            if matches!(mode, ProjectSelectMode::AllowCreateWithDefaultProjectNote)
+                && project.name == "My Project"
+            {
+                "My Project (default starter project)".to_string()
```
viadezo1er, line 126: Minor nitpick:
If you have a project named 'My Project (default starter project)', this is shown:
```
? Select project ›  
❯ + Create new project
  My Project (default starter project)
  My Project (default starter project)
```
and there is no way to distinguish the 2 if you forget they are sorted by alphabetical order.

I think this is fine and no one will name their project 'My Project (default starter project)'.

viadezo1er (approved):

src/ui/select.rs
AbhiPrasad: good catch! I fixed this so that we only render `(default starter project)` is there is a single project named `My Project`

PR #163, opened 2026-04-30
Dataset pipelines

ankrgyl:


README.md
```diff
@@ -177,6 +178,38 @@ bt eval foo.eval.ts -- --description "Prod" --shard=1/4
...
+### `bt datasets pipeline`
+
+Run full dataset pipelines declared with `DatasetPipeline(...)`, or stage fetch/transform/push.
+
+```bash
+# One-shot execution: discover refs, transform, and insert up to 100 new rows.
+bt datasets pipeline run ./pipeline.ts --limit 100
+
+# Staged execution for inspection or agent editing.
+bt datasets pipeline fetch ./pipeline.ts --limit 500
```
AbhiPrasad: why is this called fetch here, but the enum is called pull? Can we make the naming more consistent?

src/datasets/pipeline.rs
```diff
@@ -0,0 +1,2532 @@
...
+    if let Some(org_name) = source.org_name.as_deref() {
+        source_base.org_name = Some(org_name.to_string());
+    }
+    let ctx = login(&source_base).await?;
+    let client = ApiClient::new(&ctx)?;
+    let project = resolve_source_project(base, &client, source).await?;
+    Ok((ctx, client, project))
+}
+
+fn discovery_filter(
```
AbhiPrasad, line 1890: should we add a timestamp filter of some kind? Just to constrain the queries here?

src/datasets/pipeline.rs
```diff
@@ -0,0 +1,2532 @@
...
+    #[arg(long = "target-dataset")]
+    target_dataset: Option<String>,
+}
+
+#[derive(Debug, Clone, Args)]
+struct PipelineFetchOptions {
+    /// Maximum number of source refs to discover
+    #[arg(
+        long,
+        alias = "target",
```
AbhiPrasad: I would prefer if we didn't have this alias. `limit` aligns better with the rest of the CLI options, and it might be confusing that this is referring to source refs, and not final row count (while things like `--target-dataset` refer to the output).

src/datasets/pipeline.rs
```diff
@@ -0,0 +1,2532 @@
...
+
+    if is_uuid_like(dataset_name) {
+        bail!(
+            "dataset id '{}' not found in project '{}'",
+            dataset_name,
+            project.name
+        );
+    }
+
+    datasets_api::create_dataset_with_metadata(
```
AbhiPrasad, line 1825: auto creating datasets like this might be surprising behaviour. Means that folks run into issues if they make a spelling mistake or something similar. But I don't feel strongly about this, just figured being explicit about it might be better for the agents.

README.md
ankrgyl: good catch, at some point i renamed from `fetch` to `pull` and there were some lingering references.

src/datasets/pipeline.rs
ankrgyl: good catch

ankrgyl: added a `--window` argument which defaults to 1d and is always and'd in.

src/datasets/pipeline.rs
ankrgyl: agree

src/datasets/pipeline.rs
ankrgyl: i generally agree but our SDKs auto create projects and datasets so i think as-is, this is more consistent with our current semantics.

AbhiPrasad (approved):

lforst:
Did some changes in the JS version we need to check here:
- `bt` should probably just read off the `__braintrust_dataset_pipelines` global instead of `getRegisteredDatasetPipelines()` which is there or may not be there, and is an array of definitions (this is to reduce API surface in the sdk and has very little drawback here I believe)
- I feel like we should not expose the origin property to the user. It should just be filled automatically
- Changed the TS typing so that `trace` is always available in the transform input.

General Qs:
- Should we accept an `orgName` for the `target` of the definition? It seems implicit from the `source`
- Should we expose the `id` to users from the transform result? Is this something user-facing?
- In the [design doc](https://www.notion.so/braintrustdata/Dataset-pipelines-34df78580289803e801eef0ea689bc44) we were passing the `id` of the span to the `transform` funciton. In the JS implementation this doesn't seem to be the case. Could be pretty important to find a span in the trace.

ankrgyl:
Good stuff. comments inline

> Did some changes in the JS version we need to check here:
> 
> * `bt` should probably just read off the `__braintrust_dataset_pipelines` global instead of `getRegisteredDatasetPipelines()` which is there or may not be there, and is an array of definitions (this is to reduce API surface in the sdk and has very little drawback here I believe)
> * I feel like we should not expose the origin property to the user. It should just be filled automatically
> * Changed the TS typing so that `trace` is always available in the transform input.

updated this stuff. See https://github.com/braintrustdata/braintrust-sdk-python/pull/470

> 
> General Qs:
> 
> * Should we accept an `orgName` for the `target` of the definition? It seems implicit from the `source`

Yep

> * Should we expose the `id` to users from the transform result? Is this something user-facing?

https://github.com/braintrustdata/braintrust-sdk-javascript/pull/2055

> * In the [design doc](https://www.notion.so/braintrustdata/Dataset-pipelines-34df78580289803e801eef0ea689bc44) we were passing the `id` of the span to the `transform` funciton. In the JS implementation this doesn't seem to be the case. Could be pretty important to find a span in the trace.

Fixed

PR #166, opened 2026-04-30
Clean up profile semantics

ankrgyl:
If you have a project configured locally or globally, and specify `--profile`, we'd still try to use that configuration. Now, we store the profile (if it's set) in the config, and check whether or not it matches. And if it doesn't, we don't use the configured stuff.


src/auth.rs
```diff
@@ -598,14 +596,32 @@ fn resolve_api_key_override(base: &BaseArgs) -> Option<String> {
 
 pub async fn resolve_auth(base: &BaseArgs) -> Result<ResolvedAuth> {
     let mut store = load_auth_store()?;
-    let cfg_org = crate::config::load().ok().and_then(|c| c.org);
-    let mut auth = resolve_auth_from_store_with_cfg(base, &store, &cfg_org)?;
+    let mut auth_base = base.clone();
+    let cfg_org = if trimmed(base.profile.as_deref()).is_none()
```
viadezo1er: reuse trimmed_option rather than duplicating it as trimmed
``` diff
diff --git a/src/auth.rs b/src/auth.rs
index 3b1348b..146b5aa 100644
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -597,14 +597,12 @@ fn resolve_api_key_override(base: &BaseArgs) -> Option<String> {
 pub async fn resolve_auth(base: &BaseArgs) -> Result<ResolvedAuth> {
     let mut store = load_auth_store()?;
     let mut auth_base = base.clone();
-    let cfg_org = if trimmed(base.profile.as_deref()).is_none()
-        && trimmed(base.org_name.as_deref()).is_none()
+    let cfg_org = if crate::config::trimmed_option(base.profile.as_deref()).is_none()
+        && crate::config::trimmed_option(base.org_name.as_deref()).is_none()
     {
         let cfg = crate::config::load().unwrap_or_default();
-        auth_base.profile = cfg.profile.and_then(|value| {
-            let trimmed = value.trim();
-            (!trimmed.is_empty()).then(|| trimmed.to_string())
-        });
+        auth_base.profile =
+            crate::config::trimmed_option(cfg.profile.as_deref()).map(str::to_string);
         auth_base.profile.is_none().then_some(cfg.org).flatten()
     } else {
         None
@@ -805,10 +803,6 @@ where
     })
 }
 
-fn trimmed(value: Option<&str>) -> Option<&str> {
-    value.map(str::trim).filter(|value| !value.is_empty())
-}
-
 async fn run_login_set(base: &BaseArgs, args: AuthLoginArgs) -> Result<()> {
     if args.oauth {
         return run_login_oauth(base, args).await;
diff --git a/src/config/mod.rs b/src/config/mod.rs
index d4e20db..c0d8dc8 100644
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -197,7 +197,7 @@ fn config_matches_context(base: &BaseArgs, cfg: &Config, resolved_org: Option<&s
     }
 }
 
-fn trimmed_option(value: Option<&str>) -> Option<&str> {
+pub(crate) fn trimmed_option(value: Option<&str>) -> Option<&str> {
     value.map(str::trim).filter(|value| !value.is_empty())
 }
```

src/sync.rs
```diff
@@ -532,14 +532,16 @@ struct JsonlPartWriter {
 }
 
 pub async fn run(base: BaseArgs, args: SyncArgs) -> Result<()> {
```
viadezo1er, line 534: small code reuse
```diff
diff --git a/src/sync.rs b/src/sync.rs
index 4cabb26..b19db31 100644
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -532,16 +532,18 @@ struct JsonlPartWriter {
 }
 
 pub async fn run(base: BaseArgs, args: SyncArgs) -> Result<()> {
+    if let SyncCommand::Status(status) = args.command {
+        return run_status(base.json, status);
+    }
+
+    let ctx = login(&base).await?;
+    let client = ApiClient::new(&ctx)?;
+    let project = base.project.clone().or_else(|| {
+        crate::config::configured_project_for_context(&base, ctx.login.org_name().as_deref())
+    });
+
     match args.command {
         SyncCommand::Pull(pull) => {
-            let ctx = login(&base).await?;
-            let client = ApiClient::new(&ctx)?;
-            let project = base.project.clone().or_else(|| {
-                crate::config::configured_project_for_context(
-                    &base,
-                    ctx.login.org_name().as_deref(),
-                )
-            });
             run_pull(
                 base.json,
                 base.verbose,
@@ -553,17 +555,9 @@ pub async fn run(base: BaseArgs, args: SyncArgs) -> Result<()> {
             .await
         }
         SyncCommand::Push(push) => {
-            let ctx = login(&base).await?;
-            let client = ApiClient::new(&ctx)?;
-            let project = base.project.clone().or_else(|| {
-                crate::config::configured_project_for_context(
-                    &base,
-                    ctx.login.org_name().as_deref(),
-                )
-            });
             run_push(base.json, &ctx, &client, project.as_deref(), push).await
         }
-        SyncCommand::Status(status) => run_status(base.json, status),
+        SyncCommand::Status(_) => unreachable!(),
     }
 }
```

viadezo1er (commented):
I commited locally so I can just git push these changes if I haven't made a mistake.

src/switch.rs
```diff
@@ -229,7 +230,16 @@ async fn validate_or_create_project(client: &ApiClient, name: &str) -> Result<ap
...
-fn apply_switch_config(cfg: &mut config::Config, org_name: &str, project: &api::Project) {
+fn apply_switch_config(
+    cfg: &mut config::Config,
+    profile_name: Option<&str>,
+    org_name: &str,
+    project: &api::Project,
+) {
+    cfg.profile = profile_name
+        .map(str::trim)
+        .filter(|value| !value.is_empty())
```
AbhiPrasad: If I do something like

```
bt switch --profile work foo/proj1
```

and then do

```
bt switch proj2
```

then the profile becomes `null` in the saved config. Is it okay?

ankrgyl: fixed

src/sync.rs
ankrgyl: thx fixed

src/auth.rs
ankrgyl: fixed

AbhiPrasad (approved):

PR #165, opened 2026-04-30
fix: various bugs

viadezo1er:
Concurrency issue with multiple bt switch (although it seems unlikely it would have ever triggered)
Unicode issues causing thread panic (try `bt sync pull --window 1é` on the main branch)
`bt functions pull` could ask for rows with invalid project id
`bt view logs/trace` next page hint lacked the `--limit` argument
Replace manual `.tmp` file write with `tempfile::NamedTempFile`
Deduplicate `parse_duration_to_seconds`

PR #175, opened 2026-05-04
fix: unicode issues with bt sync pull

viadezo1er:
Fix the thread panic with `bt sync pull --window 1é`:
```
bt sync pull --window 1é
✔ Select project · .....
✔ Select object · .....
thread 'main' (233982) panicked at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/str/mod.rs:833:21:
byte index 2 is not a char boundary; it is inside 'é' (bytes 1..3) of `1é`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Part of https://github.com/braintrustdata/bt/pull/165

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -2607,6 +2607,34 @@ fn build_root_spans_query(
     parts.join(" | ")
 }
 
+fn parse_duration_to_seconds(input: &str) -> Result<u64> {
+    let trimmed = input.trim();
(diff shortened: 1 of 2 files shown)


AbhiPrasad (approved):

viadezo1er:
Rebased (and botched it so had to push force multiple times).

PR #182, opened 2026-05-13
add warning/alert when there are provider keys older than 6 months

daneyul:
We have this in the product UI (configs page) but the CLI should also probably surface some sort of notification when a provider key is older than 6 months. This can be displayed just once and check daily for newly stale keys

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -33,6 +33,7 @@ const KEYCHAIN_SERVICE: &str = "com.braintrust.bt.cli";
 const OAUTH_SCOPE: &str = "mcp";
 const OAUTH_CALLBACK_TIMEOUT: Duration = Duration::from_secs(300);
 const OAUTH_REFRESH_SAFETY_WINDOW_SECONDS: u64 = 60;
+const AI_PROVIDER_KEY_STALENESS_CHECK_INTERVAL_SECONDS: i64 = 24 * 60 * 60;
 static SECRET_STORE_FALLBACK_WARNED: AtomicBool = AtomicBool::new(false);
 static AI_PROVIDER_KEY_STALENESS_WARNED: AtomicBool = AtomicBool::new(false);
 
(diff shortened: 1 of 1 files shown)


viadezo1er (changes requested):
The HTTP call to check the state of the LLM API key fire on every login (and can slow it down since it's not async), which is not desirable.

Maybe add a `last_checked_at` timestamp field to the state file and check the state at most once per day?

src/auth.rs
```diff
@@ -31,7 +33,9 @@ const KEYCHAIN_SERVICE: &str = "com.braintrust.bt.cli";
 const OAUTH_SCOPE: &str = "mcp";
 const OAUTH_CALLBACK_TIMEOUT: Duration = Duration::from_secs(300);
 const OAUTH_REFRESH_SAFETY_WINDOW_SECONDS: u64 = 60;
+const AI_PROVIDER_KEY_STALENESS_CHECK_INTERVAL_SECONDS: i64 = 24 * 60 * 60;
```
daneyul, line 36: @viadezo1er checking once a day here. do you think it makes sense to show a warning only once? I think I've set this code to show a warning once to a user, and it doesn't show the same warning again if the user has not changed the key.

viadezo1er: I think it's ok, it doesn't really matter if the user changes their key after 6months or 6 months + a week because they had to be nagged about it a few times.

viadezo1er:
@daneyul Have you successfully printed the warning? (by changing the delay from 6 months to 6s for example)

daneyul:
> Daniel Nguyen (@daneyul) Have you successfully printed the warning? (by changing the delay from 6 months to 6s for example)

@viadezo1er 
CleanShot 2026-05-13 at 15 03 32@2x

viadezo1er:
I think this works but it's cleaner to use the same mechanism in bt as in the the web server (since this warning is already shown in the web ui).

daneyul:
> I think this works but it's cleaner to use the same mechanism in bt as in the the web server (since this warning is already shown in the web ui).

what do you mean?

viadezo1er (approved):

PR #188, opened 2026-05-15
feat(setup): Use new browser signup/signin and org/project selection flow

lforst:
Uses new sign-in flow introduced in
- https://github.com/braintrustdata/braintrust/pull/13858
- https://github.com/braintrustdata/braintrust/pull/13869

DO NOTE MERGE until those are merged.

Changes:
- Opens the users browser on the newly introduced wizard flow pages
- Polls the session endpoint until the user has completed it
- Picks up api key, org and project from poll result and prints selected org/project to the user

Tested locally by just running while braintrust dev server is running: `cargo run -- setup --app-url http://localhost:3000 --no-skills`

https://github.com/user-attachments/assets/9714ce08-630b-401c-b298-e9558d8094ac


viadezo1er (changes requested):
Edit: forgot `--api-url` when specifying `--app-url`, forget this issue.

~I think this breaks login for all other bt commands:~
```
bt auth login
✔ Select login method · OAuth (browser)
Opening browser for OAuth authorization...
If it does not open, visit:
https://api.braintrust.dev/oauth/authorize?response_type=code&client_id=bt_cli_default&state=...
Waiting for browser authorization...
Paste code=...&state=... if callback doesn't complete
error: request failed (401 Unauthorized): Failed to validate Clerk JWT: Error: Unable to find a signing key in JWKS that matches the kid='...' of the provided session token. Please make sure that the __session cookie or the HTTP authorization header contain a Clerk-generated session JWT. The following kid is available: ... [timestamp=1779140533.591]
```

lforst:
> I think this breaks login for all other bt commands:
> 
> ```
> bt auth login
> ✔ Select login method · OAuth (browser)
> Opening browser for OAuth authorization...
> If it does not open, visit:
> https://api.braintrust.dev/oauth/authorize?response_type=code&client_id=bt_cli_default&state=...
> Waiting for browser authorization...
> Paste code=...&state=... if callback doesn't complete
> error: request failed (401 Unauthorized): Failed to validate Clerk JWT: Error: Unable to find a signing key in JWKS that matches the kid='...' of the provided session token. Please make sure that the __session cookie or the HTTP authorization header contain a Clerk-generated session JWT. The following kid is available: ... [timestamp=1779140533.591]
> ```

I don't think I changed the code path for `bt auth login`. This looks more like an error that came from the backend tbh. Could you point to one of my changes that you think would trigger this?

viadezo1er:
Classic PEBKAC case, no issue in the code. I forgot to login incognito mode and the `--api-url` flag.

viadezo1er:
Previously `bt setup` kept the user logged-in, so they could run other braintrust commands afterwards (including another `bt setup`).
This is no longer the case with this pr.
Is it voluntary?

lforst:
> Previously `bt setup` kept the user logged-in, so they could run other braintrust commands afterwards (including another `bt setup`). This is no longer the case with this pr. Is it voluntary?

Good catch, updated it so that it saves the profile. We don't want this behavior for spark tho. Doing this just to have backwards compat. In spark we should ask the user whether they want to install the braintrust cli.

viadezo1er:
Now the `-o/--org/-p/--project` flag are useless since the user chooses the org/project in the browser, I think that's fine. The only possible thing I see this breaking would be using `bt setup` in CI, and we are probably the only ones doing this.
So maybe warrants a cleanup later but fine.

viadezo1er (approved):

PR #204, opened 2026-05-22
ci(release): publish stable releases from version bumps

AbhiPrasad:
Replace the separate tag-on-version-bump workflow with a single stable release flow driven by Cargo.toml changes on main. Operators prepare a release by running release-bump-pr with the target semver version, which opens or updates a PR that changes Cargo.toml and Cargo.lock. When that PR merges, the release workflow reads the new Cargo.toml version, creates and pushes the matching v<major>.<minor>.<patch> tag if it does not already exist, then uses cargo-dist to plan and build the release artifacts.

Publishing is gated behind the release GitHub Environment. The workflow builds artifacts before requesting approval, then only uploads the artifacts, creates the GitHub Release, and publishes installers after the deployment is approved. Manual release dispatch still supports publishing an existing semver tag directly.

The minimal smoke-install workflow now consumes the resolved release tag from a release-metadata artifact emitted by the release workflow. This keeps post-release smoke tests aligned with the exact tag the release used, including manually dispatched releases where the workflow_run branch can still be main.

Document the stable release process in CONTRIBUTING so the expected operator flow is clear.


.github/workflows/release-bump-pr.yml
```diff
@@ -0,0 +1,112 @@
...
+          for i, line in enumerate(lines):
+              if line.startswith("version = "):
+                  old = line
+                  lines[i] = f'version = "{version}"\n'
+                  break
+          if old is None:
+              raise SystemExit("failed to find version in Cargo.toml")
+          cargo_toml.write_text("".join(lines))
+
+          cargo_lock = Path("Cargo.lock")
```
viadezo1er: We could use `cargo update --workspace --offline` instead of parsing the Cargo.lock toml.
From the `cargo update` doc:
*[--workspace](https://doc.rust-lang.org/cargo/commands/cargo-update.html#option-cargo-update---workspace)
Attempt to update only packages defined in the workspace. Other packages are updated only if they don’t already exist in the lockfile. This option is useful for updating Cargo.lock after you’ve changed version numbers in Cargo.toml.*

AbhiPrasad: good call, just did that and pushed it up

viadezo1er (approved):
Unless we actively try to break things this should be good.

PR #205, opened 2026-05-26
Fix a few footguns in topics CLI

ankrgyl:
* Constrain status queries to 24h by default. Users can provide a larger window explicitly
* Don't run expensive queries in the default status check

### DIFF
--- a/src/topics/api.rs
+++ b/src/topics/api.rs
@@ -2114,13 +2114,31 @@ fn status_progress_time_filter_clause(
 ) -> Option<(String, i64)> {
     let window_seconds = match progress_window_seconds_override {
         Some(window_seconds) => window_seconds,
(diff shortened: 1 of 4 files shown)


src/topics/api.rs
homin-bt, line 2096: would it be good to keep the ability to work over absolute time ranges?

homin-bt (approved):

src/topics/api.rs
homin-bt: er, not sure what happened there, but this was supposed to point here: https://github.com/braintrustdata/bt/pull/205/changes/BASE..195dc6fb546941a9e9400ed0e64291b4dcd52304#diff-2f5ab653ff7884cb570e858b0603bdecb8119642a2777fcb16966ca7e0151a39L2096

ankrgyl: Yep

PR #206, opened 2026-05-27
Fix Python remote parameter list serialization

joshuawootonn:
## Summary

I don't think the python remote eval case was tested in the `bt` implementation of parameters. The problem was that we were trying serialize the new internal type (`RemoteEvalParameters`) I created in v0.10.0 of the python parameters implementation with the old serializer (`parameters_to_json_schema`). I updated the code to use the new serializer and everything works now. It was already working for the vanilla python SDK.
 
## Description

Customer logs showed `/list` endpoint for `bt` run remote evals was 500'ing when using saved Parameters.  

```text
[bt eval dev:list] stderr: AttributeError: 'RemoteEvalParameters' object has no attribute 'items'
...
File "/Users/divitoa/.cache/bt/eval-runners/0.8.0/eval-runner.py", line 570, in build_eval_definitions
  "parameters": parameters_to_json_schema(evaluator.parameters) if evaluator.parameters else {},
File "/Users/divitoa/code/pine/morgan/.venv/lib/python3.14/site-packages/braintrust/parameters.py", line 430, in parameters_to_json_schema
  for name, schema in parameters.items():
                    ^^^^^^^^^^^^^^^^
```

I created a reproduction (see demos) and found that `bt` run remote evals were also not working for static parameters. Instead of a 500 though, it was just showing a remote eval empty state in the playground. 

After investigating, I realized that the DTO returned from the static parameter case was different for `bt` than it was for the vanilla Python SDK. That helped me realize that this implementation wasn't taking into account [the new containerized DTO @ibolmo and I designed back](https://github.com/braintrustdata/braintrust-sdk-python/pull/90#discussion_r2955044936) when I was first working on the Parameter's API. Swapping `parameters_to_json_schema` to `serialize_remote_eval_parameters_container` was the solution, and fixed both cases by returning a DTO the playground understands from the `/list` endpoint. 

## Demo

Me going through the problem (sorry for the low quality on this video)

https://github.com/user-attachments/assets/b53b637b-3ddd-4d0a-87cc-74433c477c5e

The working thing: 

https://github.com/user-attachments/assets/d3a92c05-39f5-498e-b1a9-fd858e020e35

Old Python SDKs work in and outside of bt:

https://github.com/user-attachments/assets/cfc29948-b646-415a-8ca0-f3bb6acf1473

### DIFF
--- a/scripts/eval-runner.py
+++ b/scripts/eval-runner.py
@@ -27,6 +27,11 @@ try:
     )
     from braintrust.logger import Dataset, init as init_logger_experiment, parent_context, _internal_get_global_state
     from braintrust.parameters import parameters_to_json_schema, validate_parameters
+
+    try:
+        from braintrust.parameters import serialize_remote_eval_parameters_container
+    except ImportError:
+        serialize_remote_eval_parameters_container = None
     from braintrust.util import eprint
     from braintrust.span_identifier_v4 import parse_parent
 except Exception as exc:  # pragma: no cover - runtime guard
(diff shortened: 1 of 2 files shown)


scripts/eval-runner.py
```diff
@@ -567,12 +572,20 @@ def build_eval_definitions(evaluator_instances: list[EvaluatorInstance]) -> dict
...
             "scores": scores,
         }
     return definitions
 
 
+def serialize_evaluator_parameters(parameters: Any) -> dict[str, Any] | None:
+    if not parameters:
+        return {}
+    if serialize_remote_eval_parameters_container is not None:
+        return serialize_remote_eval_parameters_container(parameters)
```
joshuawootonn, line 585: This function was shipped with the release of the Python parameters SDK, So its presence as an import will indicate that we have the right parameter type for serializing into a containerized remote eval value.

j13huang (approved):

AbhiPrasad (approved):

PR #208, opened 2026-05-27
feat: package bt as npm pkg

viadezo1er:
~Need help to get an `NPM_TOKEN` to publish the `bt` binaries in `@braintrust` on `npm`.
This token should be a secret in the release environment.~ -> EDIT from Luca: We are using trusted publishing.

Publish the `bt` binaries to `npm`.


lforst (commented):
Thought: I vote for naming the package `@braintrust/cli`. If the binary in package.json is called `bt` and they install the package with `npm i -g @braintrust/cli`, they will still be able to invoke the cli globally with `bt`.

`@braintrust/bt` is a bit of a pleonasm. Also we already have places where we refer to `bt` as "CLI":

Screenshot 2026-05-28 at 11 38 17

lforst:
I set up the initial packages and trusted publishing for them, so this should be unblocked as long as we publish from release.yml with the release environment:

Screenshot 2026-06-02 at 13 28 24

lforst (approved):

lforst:
Actually we're using `npm-publish` as environment now because the way we set up the environmentss is annyoing to approve on github.

PR #209, opened 2026-05-27
feat: publish bt as cargo pkg to crates.io

viadezo1er:
# Keep this PR to conserve the instructions to publish bt if we need it one day.
Requires
- [ ] the rust sdk to be updated on [crates.io](https://crates.io/crates/braintrust-sdk-rust), I assume version `"0.1.0-alpha.3" but can be anything else.
- [ ] `lingua` to be published (as `braintrust-lingua`?)
  - [ ] `braintrust-serde-json` to be published. Due to https://github.com/serde-rs/json/issues/1157 and our fix https://github.com/braintrustdata/braintrust/pull/8779 we have duplicated `serde-rs` which we would need to publish too. It could be named `braintrust-serde-json` with a warning to use upstream `serde` instead since this is not for general use.
- [ ] run `cargo update` once this is done and these dependencies are added to crates.io

Would be nice to have the crate name `bt`, otherwise `braintrust-bt`?
As per https://github.com/rust-lang/crates.io/issues/326#issuecomment-3601635445 (comment from Tobias Bieniek, crates.io team co-lead), _deleting [your] GitHub account will make your [crates.io] account inaccessible and your package unmaintained_.
All of the repos/github pages/account listed on the [maintainer page](https://crates.io/users/ceram1) are 404. I believe that's why the maintainer could not transfer ownership to us (see emails in the sdk mailing list). **See https://linear.app/braintrustdata/issue/SDK-38/packaging for details about this.**
I suppose next step is contacting crates.io support.

The email of the (technically current) maintainer of the bt crate is in the cargo.toml: https://docs.rs/crate/bt/0.1.0/source/Cargo.toml, he is https://crates.io/users/ceram1.

CI doesn't pass since it's searching for the nonexistant crate/version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -488,3 +488,29 @@ jobs:
           & $btExe --version
           $env:PATH = "$binDir;$env:PATH"
           & $btExe self update --check --channel stable
+
+  publish-crate:
+    needs:
+      - plan
+      - host
+      - announce
+    if: ${{ needs.announce.result == 'success' && !fromJson(needs.host.outputs.val).announcement_is_prerelease }}
+    runs-on: ubuntu-22.04
+    environment: release
+    env:
+      CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+      BT_VERSION_STRING: ${{ needs.plan.outputs.release-version }}
+    steps:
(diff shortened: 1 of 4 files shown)


lforst (approved):
Obviously please wait for us to have the `bt` crate ownership before merging but lg

PR #210, opened 2026-05-27
Fix Python remote eval parameter serialization in dev mode

evanmkeith:
## Summary

  Fixes Python `bt eval --dev` `/list` responses for evals that define `parameters={...}`.

  The Python eval runner was emitting parameters as a raw JSON Schema object via `parameters_to_json_schema(...)`. The Braintrust UI expects remote eval parameters to use the serialized parameter container shape, such as `braintrust.staticParameters` or `braintrust.parameters`. As a result, `/list` could return `200 OK` while the UI still failed to parse the evaluator manifest and showed the generic connection/listing error.

  This updates the Python runner to serialize parameters with the same remote eval parameter container shape used by the Python SDK devserver/push flow.

  ## Test Plan

  - Add a Python list-mode fixture with a Pydantic parameter model.
  - Assert `BT_EVAL_DEV_MODE=list` emits `parameters.type = "braintrust.staticParameters"`.
  - Run:
    - `python3 -m py_compile scripts/eval-runner.py tests/evals/py/remote_list_params/eval_remote_list_params.py`
    - `cargo fmt --check`
    - `cargo test eval_python_runner_list_mode_serializes_remote_parameter_container --test eval_fixtures -- --nocapture`

### DIFF
--- a/scripts/eval-runner.py
+++ b/scripts/eval-runner.py
@@ -27,6 +27,10 @@ try:
     )
     from braintrust.logger import Dataset, init as init_logger_experiment, parent_context, _internal_get_global_state
     from braintrust.parameters import parameters_to_json_schema, validate_parameters
+    try:
+        from braintrust.parameters import serialize_remote_eval_parameters_container
+    except ImportError:
+        serialize_remote_eval_parameters_container = None
     from braintrust.util import eprint
     from braintrust.span_identifier_v4 import parse_parent
 except Exception as exc:  # pragma: no cover - runtime guard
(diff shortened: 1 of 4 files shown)


spencerseale:
This is blocking on passing params to the playground from remote evals. Subscribed for when this ships

AbhiPrasad (approved):

AbhiPrasad:
please rebase and we can get this merged in and released!

scripts/eval-runner.py
```diff
@@ -567,12 +571,47 @@ def build_eval_definitions(evaluator_instances: list[EvaluatorInstance]) -> dict
...
+            elif parameter_type == "model":
+                parameter = {"type": "model"}
+            else:
+                parameter = {"type": "data", "schema": property_schema}
+
+            if "default" in property_schema:
+                parameter["default"] = property_schema["default"]
+            if isinstance(property_schema.get("description"), str):
+                parameter["description"] = property_schema["description"]
+            static_parameters[name] = parameter
```
joshuawootonn, line 606: I'm not sure why this section is necessary. Looking to understand more.

joshuawootonn (approved):
The rest of it looks good though, and for most recent SDK installers they will not hit this branch.

joshuawootonn:
This PR is pretty similar to something I shipped last week: https://github.com/braintrustdata/bt/pull/206/

joshuawootonn (changes requested):
Ok I'm realizing that this is almost the same PR I shipped last week, which is why rebasing is necessary. https://github.com/braintrustdata/bt/pull/206

I failed to check in on what the releasing policy is for this repo / for `bt` CLI so I don't think it's shipped yet. Sorry for dropping the ball everyone!

I would like us to avoid shipping this PR. I think we just need to get a new version dropped.

evanmkeith:
already fixed [here](https://github.com/braintrustdata/bt/pull/206)

scripts/eval-runner.py
joshuawootonn: This is unnecessary. 

If you are using python SDK >= v0.10.0 `serialize_remote_eval_parameters_container` will be defined and the playground can handle the container type created for parameters

If you are using python SDK < v0.10.0 `serialize_remote_eval_parameters_container` will not be defined but `parameters_to_json_schema` won't actually create a JSON schema. In that old SDK version it was creating the object of JSON schemas that the playground also support.

So there's not a way of hitting this endpoint in which the new version of `parameters_to_json_schema` is called and returns an actual JSON schema. Which is the error that will cause an error in the playground?

PR #221, opened 2026-06-03
Dedicated CI environnement for canary releases

viadezo1er:
Currently, canary releases use the `release` env for their `AZURE` secrets.
However this env requires manual approving, which in practice block all canary releases.

This PR make canary releases use a new env, release-canary, so they can be deployed without manual approval. This env will also have the `AZURE` secrets once they exist.

## TODO before merging:
- create a `release-canary` env, restricted to `main`. No variable/secrets needed. The `AZURE`-related secrets/variables will be added later once it's ready.


AbhiPrasad (approved):
I also created a `release-canary` environment

PR #225, opened 2026-06-04
Merge parameter data updates

j13huang:
`bt functions push` currently always replaces the underlying function data for parameters, so pushing a new set of parameters overwrites any manual updates that have been made to the parameter data.

This PR makes it so that fields are updated via merge semantics. I'm also planning on reverting this behavior in the sdk 
https://github.com/braintrustdata/braintrust-sdk-javascript/pull/1363/files#diff-dc987ecde17ce424aeab43639ec1262c84c53976bdd7f644c8b8b1627c5eb0f3R629
so that by default, `bt functions push` only updates the schema. Future work will be adding some API surface area for the user to update the underlying saved parameter values themselves (https://github.com/braintrustdata/braintrust-sdk-javascript/pull/2094)

### DIFF
--- a/src/functions/push.rs
+++ b/src/functions/push.rs
@@ -47,6 +47,8 @@ const RUNNER_COMMON_SOURCE: &str = include_str!("../../scripts/runner-common.ts"
 const PYTHON_RUNNER_COMMON_SOURCE: &str = include_str!("../../scripts/python_runner_common.py");
 const PYTHON_BASELINE_DEPS: &[&str] =
     &["pydantic", "braintrust", "autoevals", "requests", "openai"];
+const PARAMETER_FUNCTION_TYPE: &str = "parameters";
+const PARAMETER_SCHEMA_MERGE_PATH: [&str; 2] = ["function_data", "__schema"];
 // Compatibility shim for existing test harnesses and eval workflows that set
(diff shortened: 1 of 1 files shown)


src/functions/push.rs
```diff
@@ -725,6 +728,32 @@ fn build_code_function_data(
...
+            .into_iter()
+            .map(|path| {
+                Value::Array(
+                    path.iter()
+                        .map(|component| Value::String((*component).to_string()))
+                        .collect(),
+                )
+            })
+            .collect(),
+        ),
```
AbhiPrasad: you should be able to use the `json!` macro here instead of constructing via `Value::Array`.

src/functions/push.rs
```diff
@@ -3144,6 +3175,44 @@ mod tests {
         assert_eq!(calculate_upload_counts(3, None), (3, 0));
     }
 
+    #[test]
+    fn parameter_function_merge_fields_are_applied() {
+        let mut object = serde_json::json!({
+            "function_type": "parameters",
+            "is_merge": false,
```
AbhiPrasad:
```suggestion
            "_is_merge": false,
```

AbhiPrasad (changes requested):

AbhiPrasad (approved):

parkerhendo (approved):

PR #233, opened 2026-06-13
chore: switch profile/org after bt auth login

viadezo1er:
After `bt auth login`, also switch the used profile similar to running `bt switch` but only for the org.
Pass `--no-profile-switch` to prevent org switching.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ Local version and pagination-key conversion helpers:
   - First prompt chooses: `OAuth (browser)` (default) or `API key`.
   - If your API key can access multiple orgs, `bt` uses a searchable picker (alphabetized) and lets you choose a specific org or no default org (cross-org mode).
   - After login, `bt` updates the active profile/org context immediately. If `--project` is set, it also switches that project; otherwise it clears any stale default project for the new login.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -289,6 +289,8 @@ Local version and pagination-key conversion helpers:
   - `bt auth login`
   - First prompt chooses: `OAuth (browser)` (default) or `API key`.
   - If your API key can access multiple orgs, `bt` uses a searchable picker (alphabetized) and lets you choose a specific org or no default org (cross-org mode).
+  - After login, `bt` updates the active profile/org context immediately. If `--project` is set, it also switches that project; otherwise it clears any stale default project for the new login.
+  - Pass `--no-profile-switch` (or set `BRAINTRUST_NO_PROFILE_SWITCH=1`) to keep the old behavior and avoid changing active profile/org/project context.
```
lforst: When would you log in but not want to switch profile?

viadezo1er: I have no idea but that's the previous behavior so I wanted to keep it.

viadezo1er: Mandatory https://xkcd.com/1172/

lforst: If we cannot think of a reason why someone might have the old behavior, I would suggest removing it entirely. Otherwise, if we do this for every change, we end up with a giant mess at some point.

viadezo1er: Reverted

lforst (approved):

PR #238, opened 2026-06-21
Fix profile precedence for orgs

ankrgyl:
We used the profile to derive the org, so if you specified an org explicitly (which happens when you run `bt init` and pick a different org than the profile), all commands would ignore it. This fixes the resolution so that org always "sticks" and beats the profile's org in precedence.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -827,19 +827,37 @@ fn resolve_api_key_override(base: &BaseArgs) -> Option<String> {
     Some(value.to_string())
 }
 
-pub async fn resolve_auth(base: &BaseArgs) -> Result<ResolvedAuth> {
-    let mut store = load_auth_store()?;
-    let mut auth_base = base.clone();
-    let cfg_org = if crate::config::trimmed_option(base.profile.as_deref()).is_none()
-        && crate::config::trimmed_option(base.org_name.as_deref()).is_none()
-    {
-        let cfg = crate::config::load().unwrap_or_default();
-        auth_base.profile =
-            crate::config::trimmed_option(cfg.profile.as_deref()).map(str::to_string);
-        auth_base.profile.is_none().then_some(cfg.org).flatten()
+fn config_auth_context(base: &BaseArgs) -> (Option<String>, Option<String>) {
+    let cfg = crate::config::load().unwrap_or_default();
+    config_auth_context_from_config(base, &cfg)
+}
+
+fn config_auth_context_from_config(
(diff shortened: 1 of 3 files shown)


viadezo1er (changes requested):
`ced-test-1` is another org.
As you can see, even when I specify the profile, the org is unchanged even though the org used should be the profile's org since no other flag is passed.
```
❯ bt auth login
....
✓ Logged in as ced8 (profile: ced8, api: https://api.braintrust.dev)

❯ bt switch
✔ Select org · ced8
✔ Select project · My Project
✓ Switched to ced8/My Project
Wrote to /Users/cedric@braintrustdata.com/.config/bt/config.json

❯ bt status
org: ced8
project: My Project
profile: ced8
user: xq9v5n+qdevohyibp54@sharklasers.com
source: /Users/cedric@braintrustdata.com/.config/bt/config.json

❯ bt status --profile ced-test-1
org: ced8
project: (unset)
profile: ced-test-1
user: Cédric Halber (cedric@braintrustdata.com)
source: /Users/cedric@braintrustdata.com/.config/bt/config.json
```

ankrgyl:
> `ced-test-1` is another org. As you can see, even when I specify the profile, the org is unchanged even though the org used should be the profile's org since no other flag is passed.
> 
> ```
> ❯ bt auth login
> ....
> ✓ Logged in as ced8 (profile: ced8, api: https://api.braintrust.dev)
> 
> ❯ bt switch
> ✔ Select org · ced8
> ✔ Select project · My Project
> ✓ Switched to ced8/My Project
> Wrote to /Users/cedric@braintrustdata.com/.config/bt/config.json
> 
> ❯ bt status
> org: ced8
> project: My Project
> profile: ced8
> user: xq9v5n+qdevohyibp54@sharklasers.com
> source: /Users/cedric@braintrustdata.com/.config/bt/config.json
> 
> ❯ bt status --profile ced-test-1
> org: ced8
> project: (unset)
> profile: ced-test-1
> user: Cédric Halber (cedric@braintrustdata.com)
> source: /Users/cedric@braintrustdata.com/.config/bt/config.json
> ```

Hm that's actually not what i would expect. Because you selected the org to be `ced8`, and the explicit intention of the change is to prefer `org` if configured over the profile. Fundamentally I think that it's somewhat ambiguous to have both `profile` (which means auth settings + org) and `org` (which is just the org). But if we have both, org should take precedence.

PR #254, opened 2026-07-02
feat: Add custom view commands (dataset views, trace views)

lforst:


Qard:
@lforst Still want to do anything with this or should I close it?

lforst:
@Qard don't close pls

PR #257, opened 2026-07-06
Replace `bt self update` with `bt update` command

parkerhendo:
Improve the discoverabilty of the native updater by making it a top-level command.

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -473,7 +473,7 @@ jobs:
         run: |
           export PATH="${XDG_BIN_HOME:-$HOME/.local/bin}:$PATH"
           bt --version | grep -i canary
-          bt self update --check
+          bt update --check
 
   smoke-install-macos:
     needs:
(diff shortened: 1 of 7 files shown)


ankrgyl (commented):
should we keep `bt self update` as an alias?

parkerhendo:
> should we keep `bt self update` as an alias?

​ah, you're right. Reimplemented to keep (but hide) the `self update` command in favor of `bt update`

viadezo1er (approved):

viadezo1er:
Maybe update `src/self_update.rs:142`: `        "self-update is only supported for installer-based installs.\ncurrent executable: {}\nif this was installed with Homebrew/apt/choco/etc, update with that package manager",` and `README.md:398`: `- Add richer channel controls for self-update (for example pinned/branch canary selection).
`


I think you wanted to change the `self-update` wording, and there's some past slop to clean (we don't support brew/apt/choco, only curl ... | sh and npm).

viadezo1er (approved):

PR #275, opened 2026-07-16
Improve auth resolution

ankrgyl:
* Quote strings with spaces in them in shell args.
* If you do `bt auth login` it writes both a profile and org to `~/.bt/config.json`. However, if you then login with a different profile/org, it reads the org value from this `config.json` and never uses the new profile's org.


viadezo1er:
Seems like the CI failures are due to github api being down.

viadezo1er (approved):

PR #287, opened 2026-07-29
Embed the coding-agent tracing daemon

Qard:
## Dependencies

Depends on merged [braintrustdata/braintrust-coding-agent-plugins#1](https://github.com/braintrustdata/braintrust-coding-agent-plugins/pull/1), pinned at its squash-merge commit `bebff4e0349fd9d2ee4a8ce1eeeafafee91ac74b`.

That daemon PR in turn depends on [braintrustdata/braintrust-sdk-rust#86](https://github.com/braintrustdata/braintrust-sdk-rust/pull/86), pinned at commit `d33e806bf6ab9548d37355f6a5098a971ef150aa`.

The plugin source cutover is intentionally deferred to stacked [braintrustdata/braintrust-coding-agent-plugins#3](https://github.com/braintrustdata/braintrust-coding-agent-plugins/pull/3), which must land only after this daemon is available in a released `bt` CLI.

## Summary

This embeds the shared coding-agent tracing daemon in `bt` under the `bt trace` command family. `bt trace setup` and `bt trace import` are the user-facing commands. The `daemon`, `hook`, `status`, and `stop` subcommands remain directly callable by plugins and operators who know their names, but are hidden from public help and usage output as implementation details.

`bt trace import codex <session-id>` and `bt trace import claude <session-id>` locate and snapshot native session transcripts in the agents' standard data directories, resolve the current `bt` profile and project, and send the reconstructed historical trace through the production Braintrust sink. The source is explicit because session IDs alone do not identify their producer. Codex snapshots are streamed once, native turn IDs and compaction history replacements are retained, and large imports drain in small bounded operation batches to avoid losing later turns to bounded delivery queues without serializing a network flush for every boundary. Claude imports preserve native turn ownership when physical record order is not timestamp-monotonic and propagate Claude API-error markers into turn and LLM errors. Daemon restart recovery remains an internal WAL operation; it may idempotently resubmit rows under stable backend row IDs, but cannot create duplicate logical spans.

The foreground daemon initializes structured logging at INFO by default, or DEBUG with `--verbose`, and emits safe request lifecycle fields such as JSON-RPC method, request ID, event type, source, and session ID. Payloads and credentials are not logged.

`bt trace setup codex` and `bt trace setup claude` use the coding agents' own plugin managers to add the official Braintrust marketplaces and install the currently published tracing plugins. Setup also creates the shared tracing settings when needed, enables tracing, and accepts `--project` for routing. It performs a non-destructive merge: every other setting, including legacy authentication and backend fields needed by currently published plugins, is preserved unchanged.

The `bt` host resolves its normal profile, OAuth, keychain, or API-key credentials for each hook invocation and transcript import, passing the resulting token, backend URLs, organization, and project to the credential-passive daemon library. Hook errors remain fail-open so tracing cannot break a Codex or Claude Code turn.

The daemon is consumed from the exact Git commit of its migration PR rather than through a local path dependency. The integration is rebased on current `bt` main and preserves the newer top-level `update` command and command routing introduced there.

CLI regression tests verify that `setup` and `import` are advertised by `bt trace --help`, while the four hidden operational commands remain directly callable. They also cover the required import source and session ID, rejection of the removed replay spelling, hook options, graceful and idempotent stop, Codex and Claude plugin-manager invocations, default and explicit project selection, and preservation of existing settings. The superseded `bt daemon` and `bt agents` paths remain rejected.

## Validation

- `cargo fmt --check`
- `cargo test --test cli`
  - all 30 CLI tests passed.
- Daemon dependency suite: 60 tests passed under all features, including multi-turn native Codex and Claude transcript imports, non-monotonic Claude record ordering and API errors, compaction-state replacement, turn-checkpoint delivery, and restart ID stability.
- Real installed Codex and Claude Code sessions completed against deterministic mock inference and satisfied their mock ingest trace scenarios.
- Manual isolated lifecycle smoke delivered deterministic Codex rows to a local mock Braintrust backend with project routing and `braintrust.plugin.codex` span-origin metadata.
- A real three-turn Claude transcript with non-monotonic record timestamps was imported into Braintrust; all three LLM spans were parented to their corresponding turns, native API failures were marked as errors, and re-import retained the same seven span IDs.

## Project selection and managed runs

- expose `bt trace run codex|claude` with invocation-local hooks and routing
- reuse an explicit or configured project for setup, run, and project-default imports
- prompt with the existing project selector when an interactive user has not selected a project
- require `--project` with actionable guidance in non-interactive mode
- persist setup profile, organization, and project selection as the typed daemon route while leaving managed-run selection process-local

Covers [SDK-128](https://linear.app/braintrustdata/issue/SDK-128/bt-trace-run) and [SDK-129](https://linear.app/braintrustdata/issue/SDK-129/add-project-selection-to-bt-trace-import-when-ambiguous).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -488,11 +488,11 @@ dependencies = [
  "chrono",
  "crossbeam",
  "futures",
- "indexmap 2.13.0",
+ "indexmap 2.14.0",
  "regex",
  "reqwest",
  "serde",
- "serde_json 1.0.149",
+ "serde_json 1.0.151",
  "serde_repr",
  "thiserror 1.0.69",
  "tokio",
@@ -594,7 +594,7 @@ dependencies = [
 [[package]]
 name = "bt-daemon"
 version = "0.1.0"
-source = "git+https://github.com/braintrustdata/braintrust-coding-agent-plugins?rev=96f452780b4c3b1aea2325853817721e777d3c48#96f452780b4c3b1aea2325853817721e777d3c48"
+source = "git+https://github.com/braintrustdata/braintrust-coding-agent-plugins?rev=30e4b53f373940ea26c7ca943bd56c51fa431e13#30e4b53f373940ea26c7ca943bd56c51fa431e13"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -603,9 +603,9 @@ dependencies = [
  "clap",
  "regex",
  "serde",
- "serde_json 1.0.149",
+ "serde_json 1.0.151",
  "sha2",
- "thiserror 2.0.18",
+ "thiserror 2.0.19",
  "tokio",
  "tracing",
  "uuid",
@@ -1871,12 +1871,6 @@ name = "lazy_static"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
-
-[[package]]
-name = "leb128fmt"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09edd9e8b54e49e587e4f6295a7d29c3ea94d469cb40ab8ca70b288248a81db2"
 
 [[package]]
 name = "libc"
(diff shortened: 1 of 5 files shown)


viadezo1er (approved):

AbhiPrasad (approved):

AbhiPrasad (approved):

viadezo1er (approved):
Seems like `--json` works.

PR #305, opened 2026-08-13
feat: Go eval support

delner:
## What

Adds `Go` to `EvalLanguage`, so `bt eval` can run Go evals.

## Why

The Braintrust Go SDK now implements the runner protocol (`evalrunner` package),
but `bt` could not dispatch to it: `EvalLanguage` was a closed
`{JavaScript, Python}` enum and `detect_eval_language` hard-errored on any other
extension. This is the companion change.

## How

Go needs a shape `bt` did not have. For JS and Python, **bt supplies the
program** — its own embedded `eval-runner.{ts,py}` — and the user supplies files
that program imports at runtime. Go has no runtime loading: **the user's
compiled package is the runner**. So bt simply
executes what the user built. Go's compilation unit is also a *directory*, not a
file, so all inputs must resolve to one package.

```bash
bt eval --language go ./cmd/evals            # compile and run the package
bt eval ./cmd/evals                          # inferred from a *_eval.go marker
bt eval --language go --runner ./bin/evals   # run a prebuilt binary
bt eval --dev ./cmd/evals                    # dev server for the playground
```

- **Discovery**: `**/*_eval.go`, a suffix mirroring Go's own `_test.go`
  convention. The matched file is only a marker — `go run` compiles the whole
  package. Pass `--language go` for a package with no such file.
- **Command**: `go run <package dir>`. With `--runner`, the binary is spawned
  with **no arguments** — the Go runner takes all input from `BT_EVAL_*` and
  ignores argv, and appending files would break a `main` that parses flags.
- **Toolchain**: `BT_EVAL_GO_BIN` / `BT_EVAL_GO`, then `GOROOT`, then `PATH`.

Everything else — env vars, the SSE socket, the stdout manifest, `drive_eval_runner` —
is language-agnostic and unchanged.

## Other behavior changed

Worth calling out, since these are not Go-specific:

1. **`prepare_eval_runners` is now lazy.** It was writing *both*
   `eval-runner.ts` and `eval-runner.py` to disk on every spawn, before bt knew
   which language it was running — and under `--dev` that is once per HTTP
   request. It now runs inside the language arms that need it.
2. **`vendor` added to `DEFAULT_EXCLUDE_DIRS`**, so bare `bt eval` in a Go repo
   does not pick up a vendored dependency's eval files. This also applies to
   JS/Python discovery, but only in the zero-argument case.
3. **`**/*_eval.go` added to `DEFAULT_EVAL_GLOBS`.** A polyglot repo containing
   both TS evals and a `*_eval.go` file will now get "Mixed eval file types are
   not supported yet" from bare `bt eval` — an existing failure class, new
   trigger.
4. **`expand_eval_file_globs` now takes the `--language` override**, because it
   runs *before* `detect_eval_language` and would otherwise reject a Go package
   directory before the override was ever consulted.

## Watch mode

A Go binary cannot report its own sources, so the watch set is derived
statically: the package directory (whose mtime changes when files are added or
removed) plus its buildable `.go` files. Imports in other packages are not
followed.

## Tests

17 new tests. `detect_eval_language` had no coverage at all before this, so Go
adds the first. Verified end to end against the Go SDK: `/list`, a streaming
playground run, and a CLI `bt eval` run.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -27,6 +27,7 @@ use tokio::sync::mpsc;
 
 use crate::args::BaseArgs;
 use crate::auth::resolved_runner_env;
+use crate::go_runner;
 use crate::js_runner;
 use crate::python_runner;
 use crate::runner_sse;
@@ -241,6 +242,7 @@ const PY_RUNNER_SOURCE: &str = include_str!("../scripts/eval-runner.py");
 const JS_RUNNER_FIRE_AND_FORGET_ENTRY: &str =
     "\nmain().catch((err) => {\n  console.error(err);\n  process.exit(1);\n});\n";
 const PYTHON_INTERPRETER_ENV_OVERRIDES: &[&str] = &["BT_EVAL_PYTHON_RUNNER", "BT_EVAL_PYTHON"];
+const GO_TOOLCHAIN_ENV_OVERRIDES: &[&str] = &["BT_EVAL_GO_BIN", "BT_EVAL_GO"];
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq, ValueEnum)]
 pub enum EvalLanguage {
@@ -248,6 +250,8 @@ pub enum EvalLanguage {
     JavaScript,
     #[value(alias = "py")]
     Python,
+    #[value(alias = "golang")]
+    Go,
 }
 
 #[derive(Debug, Clone, Args)]
@@ -256,6 +260,8 @@ Examples:
   bt eval my.eval.ts
   bt eval --no-send-logs --runner tsx my.eval.ts
   bt eval --language python my_eval.py
+  bt eval --language go ./cmd/evals
+  bt eval --language go --runner ./bin/evals
 ")]
 pub struct EvalArgs {
     /// Eval files, directories, or glob patterns to execute (e.g. foo.eval.ts, tests/, "**/*.eval.ts").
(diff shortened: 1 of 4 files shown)


src/eval.rs
```diff
@@ -2192,6 +2215,10 @@ const DEFAULT_EVAL_GLOBS: &[&str] = &[
     "**/*.eval.cjs",
     "**/*.eval.py",
     "**/eval_*.py",
+    // Go marks special files with a suffix (`_test.go`), so evals follow the
+    // same convention. The matched file is only a marker: `go run` compiles the
+    // whole package directory regardless.
+    "**/*_eval.go",
```
AbhiPrasad, line 2221: are we sure about this approach? I guess the alternative is the special named directory.

AbhiPrasad (commented):
I think this should be using https://pkg.go.dev/cmd/go/internal/list `go list -json <package>`, to figure out what to call `go run` with

Qard:
Looks like this also needs a rebase.

@delner still working on this? I could pick this up if you've got other priorities.

delner:
Addressing your feedback @AbhiPrasad : I think we can make what I have here a little better.

src/eval.rs
delner: Changed this: now you just target the directory, no special suffix.

delner:
@AbhiPrasad your feedback should be addressed, if you want to take another look!

AbhiPrasad (approved):

Qard (approved):

PR #318, opened 2026-08-14
chore(ci): update the node used in setup node action from 20/22 to 24

viadezo1er:
Test suite now runs with node 20 & 22 & 24 instead of 20 & 22

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -517,9 +517,9 @@ jobs:
             --dir archives
           ls -la archives
 
-      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: "22.15.0"
+          node-version: "24"
           registry-url: "https://registry.npmjs.org"
       - name: Use npm 11.12.1
         run: npm install -g npm@11.12.1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: ["20", "22"]
+        node-version: ["20", "22", "24"]
     permissions:
       contents: read
     env:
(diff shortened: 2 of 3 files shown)


lforst (approved):

.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: ["20", "22"]
+        node-version: ["22", "24"]
```
AbhiPrasad: why not also do 20 here?

viadezo1er: https://github.com/braintrustdata/bt/actions/runs/32050777654/job/95449350633
Run actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

viadezo1er: Unless we use ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true, Node 20 isn't supported anymore, and I don't see a reason why we should. Maybe I'm wrong?

AbhiPrasad: this is for https://github.com/braintrustdata/bt/blob/9a815ca26747ed1f270e3eb7733adbf51b89fbda/.github/workflows/tests.yml#L61-L63

which takes in any node version. Not related to actions using a specific node version

viadezo1er: Woops
Added node 20

.github/workflows/release.yml
```diff
@@ -517,9 +517,9 @@ jobs:
             --dir archives
           ls -la archives
 
-      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: "22.15.0"
+          node-version: "24.19.0"
```
AbhiPrasad: why is this 24.19.0 and above is just node-version: "24"

can't we make this 24 as well?

viadezo1er: I didn't think deeply about it, I saw a full version "22.15.0", I kept a full version.
Now it's "24" only.

AbhiPrasad (approved):

